PR #118, opened 2015-08-13
removed camel case in name, so use "import spiceypy"

AndrewAnnex:
did some additional refactoring in setup.py


AndrewAnnex:
this pr bumps the version up to 0.6.1

PR #130, opened 2016-01-19
SpiceyPy-115

AndrewAnnex:
* Attempts to handle default string length args for ease of use


AndrewAnnex:
merged in a4950d247d8a227963a536751f3e960313d344f7

PR #158, opened 2016-07-26
added tests for

AndrewAnnex:
fovray, fovtrg, pxfrm2


AndrewAnnex:
looks like kernels are failing to download...

PR #170, opened 2016-10-26
Use tarfile to extract cspice

migueldvb:


migueldvb:
Sorry, it seems that the compression is not supported by tarfile...

AndrewAnnex:
yeah I think I ran into this awhile back hence the subprocess call. I will welcome a pr for the spelling fixes though.

PR #171, opened 2016-10-26
Fix typos in cspice script

migueldvb:


AndrewAnnex:
hey I do want to merge this in, but I have been thinking of adding a CLA to the repo as that is generally the right thing to do. It might take me a few days to get to that.

AndrewAnnex:
I may instead adopt a DCO, haven't quite learned enough yet to make a decision though.

Or I can just kick that can down the road...

AndrewAnnex:
opting to kick the proverbial can, I may adopt a DCO later though. Will merge this soon

Issue #176, opened 2017-01-07
Travis builds failing


AndrewAnnex:
During a scheduled build, all travis linux builds failed downloading CSPICE from the naif servers returning urlexceptions of some kind (see https://travis-ci.org/AndrewAnnex/SpiceyPy/jobs/189723324)

I was unable (yet) to reproduce the failure locally or through appveyor builds (both appear normal). 

Not enough information as of yet to explain issue

AndrewAnnex:
a recent build (https://travis-ci.org/AndrewAnnex/SpiceyPy/jobs/190395071) shows SSL errors
```
Download failed with URLError: <urlopen error [SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:645)>,
```

AndrewAnnex:
it looks like this might be a know issue https://github.com/kennethreitz/requests/issues/2022

but most of the users there are talking about macOS and python 2.7 or older, and I have builds across all versions failing..

AndrewAnnex:
fixed in #180

PR #179, opened 2017-01-15
reverting https links to http until I learn more about ssl/TLS to fix…

AndrewAnnex:
… the failed ssl handshake errors


AndrewAnnex:
should fix #176, but I am not happy with this as a fix

PR #202, opened 2017-09-03
Modified getspice/setup to support HTTPS connections to NAIF server

jdiazdelrio:
Hi Andrew,
I have modified the getspice/setup modules to support HTTPS connections to NAIF server with
Python environments linked to old OpenSSL libraries by using urllib3 when the OpenSSL library
version linked to the standard ssl library of python is older than OpenSSL 1.0.1g.  If it is newer, it
will follow the same approach as you had on your original code.

Only in the event of needing to user urllib3, the getspice module will install using a pip command
the required dependencies.

Unfortunately, when I realized that I should have used a branch instead of the master, it was too
late, so this Pull Request goes on my own master branch.  Sorry about this.

Cheers,
  Jorge

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project tries to adhere to [Semantic Versioning](http://semver.org/).
   than OpenSSL 1.0.1g.
 - getspice module provides now a class that handles the downloading and
   unpacking of N066 CSPICE distribution.
-- Updated setup to packe the CSPICE installation code into a class that
+- Updated setup to pack the CSPICE installation code into a class that
   extends the setuptools.command.install command.
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,16 +5,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project tries to adhere to [Semantic Versioning](http://semver.org/).
 
 ## [2.0.1] - UNRELEASED
-### Changed [JDR - ODC Space] - 2.0.1.dev1
+### Changed
 - Updated getspice module to use urllib3 for OpenSSL library versions older
   than OpenSSL 1.0.1g.
 - getspice module provides now a class that handles the downloading and
   unpacking of N066 CSPICE distribution.
-- Updated setup to include certifi as part of the required packages and
-  packed the CSPICE installation code into a class that extends the
-  setuptools.command.install command.
-- tested on Python 2.7, 3.4, 3.5 and 3.6 (Mac OS 10.12.6, Windows-7 Ultimate,
-  Cygwin_64, Cygwin_32 and Ubuntu 17.04 Desktop)
+- Updated setup to packe the CSPICE installation code into a class that
+  extends the setuptools.command.install command.
 
 ### Fixed
 - added six and numpy to setup_requires setup.py kwargs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project tries to adhere to [Semantic Versioning](http://semver.org/).
 
 ## [2.0.1] - UNRELEASED
+### Changed [JDR - ODC Space] - 2.0.1.dev1
+- Updated getspice module to use urllib3 for OpenSSL library versions older
+  than OpenSSL 1.0.1g.
+- getspice module provides now a class that handles the downloading and
+  unpacking of N066 CSPICE distribution.
+- Updated setup to include certifi as part of the required packages and
+  packed the CSPICE installation code into a class that extends the
+  setuptools.command.install command.
+- tested on Python 2.7, 3.4, 3.5 and 3.6 (Mac OS 10.12.6, Windows-7 Ultimate,
+  Cygwin_64, Cygwin_32 and Ubuntu 17.04 Desktop)
+
 ### Fixed
 - added six and numpy to setup_requires setup.py kwargs
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,16 +5,20 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project tries to adhere to [Semantic Versioning](http://semver.org/).
 
 ## [2.0.1] - UNRELEASED
+### Added
+- Completed wrapping of all needed new N66 functions
+
+### Fixed
+- added six and numpy to setup_requires setup.py kwargs
+
 ### Changed
+- changed naming of vectorToList to cVectorToPython
 - Updated getspice module to use urllib3 for OpenSSL library versions older
   than OpenSSL 1.0.1g.
 - getspice module provides now a class that handles the downloading and
   unpacking of N066 CSPICE distribution.
 - Updated setup to pack the CSPICE installation code into a class that
   extends the setuptools.command.install command.
-
-### Fixed
-- added six and numpy to setup_requires setup.py kwargs
 
 ## [2.0.0] - 2017-06-09
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project tries to adhere to [Semantic Versioning](http://semver.org/).
 ## [2.0.1] - UNRELEASED
 ### Added
 - Completed wrapping of all needed new N66 functions
+- added 3.6 classifier
 
 ### Fixed
 - added six and numpy to setup_requires setup.py kwargs
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -15,6 +15,9 @@ requirements:
         - six
         - numpy
         - certifi
+        - cffi
+        - cryptography
+        - idna
         - pyOpenSSL
         - urllib3
     run:
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -20,6 +20,7 @@ requirements:
         - setuptools
         - six
         - numpy
+        - certify
 
 test:
     imports:
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -14,6 +14,9 @@ requirements:
         - setuptools
         - six
         - numpy
+        - certifi
+        - pyOpenSSL
+        - urllib3
     run:
         - python
         - setuptools
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -13,20 +13,25 @@ requirements:
         - python
         - setuptools
         - six
-        - numpy
+        - numpy >=1.12
+        - certifi >=2017.1.23
+        - pytest
     run:
         - python
         - setuptools
         - six
-        - numpy
+        - numpy >=1.12
+        - certifi >=2017.1.23
+        - pytest
 
 test:
     imports:
         - spiceypy
     requires:
-        - numpy
+        - numpy >=1.12
         - pytest
         - six
+        - certifi >=2017.1.23
 
 about:
     home: https://github.com/AndrewAnnex/SpiceyPy
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -25,9 +25,6 @@ requirements:
         - setuptools
         - six
         - numpy
-        - certifi
-        - pyopenssl
-        - urllib3
 
 test:
     imports:
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -18,7 +18,7 @@ requirements:
         - cffi
         - cryptography
         - idna
-        - pyOpenSSL
+        - pyopenssl
         - urllib3
     run:
         - python
@@ -26,7 +26,7 @@ requirements:
         - six
         - numpy
         - certifi
-        - pyOpenSSL
+        - pyopenssl
         - urllib3
 
 test:
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -14,13 +14,11 @@ requirements:
         - setuptools
         - six
         - numpy
-        - certifi
     run:
         - python
         - setuptools
         - six
         - numpy
-        - certify
 
 test:
     imports:
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: spiceypy
-  version: "2.0.1.dev0"
+  version: "2.0.1.dev2"
 
 source:
   path: ../
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -14,6 +14,7 @@ requirements:
         - setuptools
         - six
         - numpy
+        - certifi
     run:
         - python
         - setuptools
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -14,12 +14,6 @@ requirements:
         - setuptools
         - six
         - numpy
-        - certifi
-        - cffi
-        - cryptography
-        - idna
-        - pyopenssl
-        - urllib3
     run:
         - python
         - setuptools
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -25,6 +25,9 @@ requirements:
         - setuptools
         - six
         - numpy
+        - certifi
+        - pyOpenSSL
+        - urllib3
 
 test:
     imports:
--- a/appveyor/meta.yaml
+++ b/appveyor/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: spiceypy
-  version: "2.0.1.dev2"
+  version: "2.0.1.dev0"
 
 source:
   path: ../
--- a/getspice.py
+++ b/getspice.py
@@ -1,9 +1,7 @@
-#!/usr/bin/env python3
-
 """
 The MIT License (MIT)
 
-Copyright (c) [2015-2017] [Andrew Annex]
+Copyright (c) [2017] [ODC Space]
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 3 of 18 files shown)


AndrewAnnex:
Thanks for the pr, not sure what you mean by the last note, just make sure your master branch has been updated to look like my master branch and that should be fine. Given that this is the first corporation sponsored contribution, I will be creating a CLA to be signed and I will also be pretty strict about licensing. I started drafting the CLA and it should be up in a few days. I will review the state of the PR as it is now, but I can see that a number of builds on travis and appveyor have failed, which will have to be fixed.

Thanks -Andrew

CHANGELOG.md
```diff
@@ -5,6 +5,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project tries to adhere to [Semantic Versioning](http://semver.org/).
 
 ## [2.0.1] - UNRELEASED
+### Changed [JDR - ODC Space] - 2.0.1.dev1
```
AndrewAnnex: I think this breaks the formatting expected by http://keepachangelog.com/, mention the contributor within the list of things changes instead of in the heading. We might as well add an Authors markdown file to this PR also.

getspice.py
```diff
@@ -1,9 +1,7 @@
-#!/usr/bin/env python3
-
 """
 The MIT License (MIT)
 
-Copyright (c) [2015-2017] [Andrew Annex]
+Copyright (c) [2017] [ODC Space]
```
AndrewAnnex: no, this goes against the terms of the  MIT license. You need to append the ODC Space license to the end of the lists here (aka after the Philipp Rasch part), leaving 3 licenses in the header.

getspice.py
```diff
@@ -22,135 +20,222 @@
...
-    print('Best option:', result.split('/')[0])
-    print('URL: ', root_url + result)
-
-    ### DOWNLOAD AND UNPACK BEST PACKAGE ###
-    root_dir = os.path.realpath(os.path.dirname(__file__))
-
-    print('\nDownloading...')
-
-    attemptSpiceDownloadXTimes(10, root_url, result, root_dir)
+__author__ = 'JorgeDiazdelRio'
```
AndrewAnnex: similar to above, and the author attribute isn't really useful to begin with so we should just not have it

getspice.py
```diff
@@ -22,135 +20,222 @@
...
+
+
+# =============================================================================
+# Handling of CSPICE downloads from HTTPS
+# ---------------------------------------
+#    Some Python distributions may be linked to an old version of OpenSSL
+#    which will not let you connect to NAIF server due to recent SSL cert
+#    upgrades on the JPL servers.  Moreover, versions older than
+#    OpenSSL 1.0.1g are known to contain the 'the Heartbleed Bug'.  Therefore
```
AndrewAnnex: Great explanation!  👍

getspice.py
```diff
@@ -22,135 +20,222 @@
...
+def _download_ssl(url):
+    """
+    Returns the CSPICE package located at the NAIF server using the input
+    ``url``.  This fuction is implemented using urllib3 and pyOpenSSL in
+    order to resolve the known issue with OpenSSL MacOS Python
+    distributions before Python3.6.x
+
+    Parameters
+    ----------
+    url: str
```
AndrewAnnex: could you reformat the style of the docstring to match the RST/sphinx format I use for the rest of spiceypy?

getspice.py
```diff
@@ -22,135 +20,222 @@
...
+    pip.main(['install', 'urllib3[secure]'])
+
+    # Force urllib3 to use pyOpenSSL
+    import urllib3.contrib.pyopenssl
+    urllib3.contrib.pyopenssl.inject_into_urllib3()
+
+    import certifi
+    import urllib3
+
+    def _download_ssl(url):
```
AndrewAnnex: ok overall I like the changes for this file, but I would rather just see one definition for `_download_ssl()`, the switch for using pyopenssl should occur within that function definition. I would just make it all a member functions of GetCSPICE.

getspice.py
```diff
@@ -22,135 +20,222 @@
...
+        package to the self._local io.ByteIO stream.
+        """
+        self._local = _download_ssl(self._rcspice)
+
+    def _unpack(self):
+        """Unpacks the CSPICE package on the given root directory. Note that
+        Package could either be the zipfile.ZipFile class for Windows platforms
+        or tarfile.TarFile for other platforms.
+
+        Note
```
AndrewAnnex: not much of reason to keep the notes in the file, as long as the licenses and git history  are kept intact.

getspice.py
```diff
@@ -22,135 +20,222 @@
...
+        -------
+        io.BytesIO
+           CSPICE package encapsulated within a io.BytesIO stream.
+        """
+        # Create a PoolManager
+        https = urllib3.PoolManager(cert_reqs='CERT_REQUIRED',
+                                    ca_certs=certifi.where())
+
+        # Send the request to get the CSPICE package.
+        response = https.request('GET', url, timeout=urllib3.Timeout(10))
```
AndrewAnnex: I had some code that did download retries that I think is worth keeping around...

requirements.txt
```diff
@@ -2,4 +2,13 @@ numpy>=1.8.0
...
 pytest>=2.9.0
 coverage>=4.0
-coveralls>=1.1
\ No newline at end of file
+coveralls>=1.1
+
+#########################################################
+# Packages required for running getspice                #
+#########################################################
+certifi                # Required by getspice
```
AndrewAnnex: add minimal versions for certifi and pyopenssl unless there is a good reason not to.

setup.py
```diff
@@ -21,8 +21,10 @@
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
+from __future__ import print_function
```
AndrewAnnex: hmm, ok I never bothered importing this before. maybe six does it implicitly, or python 2.7 just was able to deal with the python3 syntax as I had it

setup.py
```diff
@@ -43,8 +45,20 @@
...
+
+
+# Set the CPPFLAGS and LDFLAGS environment variables before running the
+# SpiceyPy setup. This is required in order to fix issues with OpenSSL
+# on Mac OSX. Further information:
+# https://cryptography.io/en/latest/installation/#building-cryptography-on-macos
+# cryptograpy is required in order to run the 'download' method of the
+# GetCSPICE class of the getspice.py module, and it needs to be build
+# before if it is not preinstalled.
+# if host_OS == 'Darwin':
```
AndrewAnnex: commented out code here...

setup.py
```diff
@@ -64,187 +78,193 @@ class BinaryDistribution(Distribution):
...
+class InstallSpiceyPy(install):
+    """Class that extends the install command and encapsulates the
+    process for installing the required CSPICE distribution at the
+    right place.
+    """
+
+    def run(self):
+        self.check_for_spice()
+    
+        print("Host OS: {}".format(host_OS))
```
AndrewAnnex: think that the format without an index number is python 3 specific, or something like that (this why every usage I have includes the number)

setup.py
```diff
@@ -64,187 +78,193 @@ class BinaryDistribution(Distribution):
...
-        test_suite='spiceypy.tests.test_wrapper.py',
-        extras_require={'testing': ['pytest']}
-    )
-finally:
-    pass
-
+            self.move_to_root_directory()
+
+    def windows_method(self):
+        if host_OS == "Windows":
```
AndrewAnnex: if statement here is not that helpful at this point

setup.py
```diff
@@ -64,187 +78,193 @@ class BinaryDistribution(Distribution):
...
+        "Development Status :: 4 - Beta",
+        "Natural Language :: English",
+        "Topic :: Scientific/Engineering",
+        "Topic :: Scientific/Engineering :: Astronomy",
+        "License :: OSI Approved :: MIT License",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3.3",
+        "Programming Language :: Python :: 3.4",
+        "Programming Language :: Python :: 3.5",
+        "Operating System :: MacOS :: MacOS X",
```
AndrewAnnex: oh since we are editing this go ahead and add 3.6 to the classifiers

jdiazdelrio:
Thanks for the review.  I will work on the comments right away.   Regarding the failures we see on Travis and Appveyor; they are somehow related to the configuration of the environments.  I'm not familiar with either environment.

Could you guide me a little bit? 

Thanks, Jorge

AndrewAnnex:
Here are some relevant links:
https://docs.travis-ci.com/
https://www.appveyor.com/docs/

The best thing to do would be to just read the build logs for some of the failing builds to determine the issue. Once you think you have a fix locally, push a new commit and the services will start a new build. 

To get to the logs just click the links that are shown after clicking the most recent red X

AndrewAnnex:
So I was looking into the CLA issue and I think I am close to a solution. The problem is that there is no simple CLA that is appropriate for MIT Licensed projects. The good news is that the GitHub   [terms of service](https://help.github.com/articles/github-terms-of-service/#6-contributions-under-repository-license ) explicitly state that unless otherwise superseded with a CLA the project's license applies implicitly to all contributions. So in short I probably don't need a CLA.

This should mean that license wise this pr is fine, including an explicit MIT license from ODC should be an even better case and it should also solve the patent concern as well. 

A CLA could still be added that just points back to the github terms of service to make sure there is an additional record of agreement.

jdiazdelrio:
All testing pass except for the AppVeyor with Python3.4 due to some issue when trying to install pysocks, which is required for urllib3.  Any ideas?

AndrewAnnex:
Maybe try using the requests library instead of urllib3. Even if it is a higher level interface to urllib3 they might have figured out the dependency issues.


-Andrew Annex

jdiazdelrio:
I have managed to fix the issue for conda, by removing any dependencies to urllib3 unless they are strictly necessary, i.e. only for distributions where OpenSSL is < 1.0.1g (Mac + Python < 3.5). Everything looks good now.

AndrewAnnex:
hey I just merged in #220 so that is why there are conflicts, but it would good to see this pr keep the functionality for FreeBSD as well

getspice.py
```diff
@@ -43,114 +41,212 @@
...
+           upgrades on the JPL servers.  Moreover, versions older than
+           OpenSSL 1.0.1g are known to contain the 'the Heartbleed Bug'.
+           Therefore this method provides two different implementations for the
+           HTTPS GET call to the NAIF server to download the required CSPICE
+           distribution package.
+        """
+        # Use urllib3 (based on PyOpenSSL).
+        if ssl.OPENSSL_VERSION < 'OpenSSL 1.0.1g':
+            import pip
+            pip.main(['install', 'pyOpenSSL'])
```
AndrewAnnex: ok, so when I first saw this, I was not a huge fan of runtime invocations of pip like this. But I thought of an alternative solution that may be worth trying. How about we just have a runtime check (the if statement above this line) and have it update the `DEPENDENCIES` list to include urllib3 and pyOpenSSL? This way, we would not need to import pip and the runtime installations could be removed. ie at the top of setup.py have a code block like: 
```python
DEPENDENCIES = ['numpy>=1.8.0', 'six>=1.9.0']
if ssl.OPENSSL_VERSION < 'OpenSSL 1.0.1g':
    DEPENDENCIES.append('pyOpenSSL>=17.3.0')
    DEPENDENCIES.append('urllib3[secure]>=1.22')
```

jdiazdelrio:
@AndrewAnnex, Could you restart from this PR the two jobs that failed.  I think there was an issue with travis. Thanks!

jdiazdelrio:
@AndrewAnnex I don't understand why the build shows as failed.  I have to re-run on my fork twice the Mac OS Python 2.7.13 tests since there was some issue with Travis. The second time it worked.  Could you please re-run it from the master branch?

latest_build

AndrewAnnex:
the builds time out occasionally, from my observation it depends on the total load on their macOS worker pool as build times increase when all of the available workers are busy.

getspice.py
```diff
@@ -43,116 +41,210 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
 
-    attemptSpiceDownloadXTimes(10, root_url, result, root_dir)
+"""
+from __future__ import print_function
```
AndrewAnnex: I think this can be safely removed

requirements.txt
```diff
@@ -2,4 +2,7 @@ numpy>=1.8.0
 six>=1.9.0
 pytest>=2.9.0
 coverage>=4.0
-coveralls>=1.1
\ No newline at end of file
+coveralls>=1.1
+certifi>=2017.7.27.1
```
AndrewAnnex: I think we can remove these additions to the requirements.txt as they are optional/dynamic depending on the ssl version. We can alternatively have them in a new file `optional-requirements.txt`.

setup.py
AndrewAnnex: same as above comment, I don't think it is needed

setup.py
```diff
@@ -45,8 +48,15 @@
 lib_dir = os.path.join(cspice_dir, 'lib')
 
 TEST_DEPENDENCIES = ['numpy>=1.8.0', 'pytest>=2.9.0', 'six>=1.9.0']
-DEPENDENCIES      = ['numpy>=1.8.0', 'six>=1.9.0']
-REQUIRES          = ['numpy', 'six']
+DEPENDENCIES = ['certifi>=2017.7.27.1', 'numpy>=1.8.0', 'six>=1.9.0']
```
AndrewAnnex: certifi should be moved to the if statement block below unless there is something I am missing

setup.py
```diff
@@ -66,189 +76,192 @@ class BinaryDistribution(Distribution):
...
+            self.move_to_root_directory()
+
+
+readme = open('README.rst', 'r')
+readmetext = readme.read()
+readme.close()
+
+setup(
+    name='spiceypy',
+    version='2.0.1.dev2',
```
AndrewAnnex: why did the version number change?

setup.py
```diff
@@ -66,189 +76,192 @@ class BinaryDistribution(Distribution):
...
+        "Operating System :: POSIX :: Linux",
+        "Operating System :: POSIX :: FreeBSD",
+        "Operating System :: Microsoft :: Windows"
+    ],
+    packages=find_packages(exclude=["*.tests"]),
+    include_package_data=True,
+    zip_safe=False,
+    distclass=BinaryDistribution,
+    package_data={'': ['*.so', "*.dll"]},
+    setup_requires=[],
```
AndrewAnnex: I think this should be set back to `DEPENDENCIES`, unless there is a reason not to

getspice.py
```diff
@@ -43,116 +41,210 @@
...
+            attempts = 10   # Let's try a maximum of attempts for getting SPICE
+            while attempts:
+                attempts -= 1
+                try:
+                    self._download()
+                except RuntimeError as error:
+                    print("Download failed with URLError: {0}, trying again after "
+                          "15 seconds!".format(error))
+                else:
+
```
AndrewAnnex: there are a lot of new lines... its a style thing but for me this is too sparse.

getspice.py
```diff
@@ -43,116 +41,210 @@
...
+            # Download the file
+            print('Downloading CSPICE for {0}...'.format(distribution))
+            attempts = 10   # Let's try a maximum of attempts for getting SPICE
+            while attempts:
+                attempts -= 1
+                try:
+                    self._download()
+                except RuntimeError as error:
+                    print("Download failed with URLError: {0}, trying again after "
+                          "15 seconds!".format(error))
```
AndrewAnnex: I would like to see the sleep call added back.

appveyor/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: spiceypy
-  version: "2.0.1.dev0"
+  version: "2.0.1.dev2"
```
AndrewAnnex: should be dev0

getspice.py
```diff
@@ -17,7 +15,7 @@
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
```
AndrewAnnex: ehh, this is picky but the unhyphenated version is what is in the original license

getspice.py
```diff
@@ -43,116 +41,210 @@
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
-FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
```
AndrewAnnex: ehh, this is picky but the unhyphenated version is what is in the original license

AndrewAnnex:
ok I left a number of comments for you to review, this pr is getting close!

getspice.py
jdiazdelrio: Ok. Changed back.

getspice.py
jdiazdelrio: Ok. Changed back

appveyor/meta.yaml
jdiazdelrio: Ok. Changed

setup.py
jdiazdelrio: There are two reasons for not setting it back to DEPENDENCIES. The second one is that setup does not require any of those packages to be installed. Now it deals with a class that imports the required modules only if needed and when install is called. The second one is that they do not play nicely together, particularly for numpy and cryptography (or any other package that may require compiling).

setup.py
jdiazdelrio: Done

getspice.py
jdiazdelrio: Done

getspice.py
jdiazdelrio: Done

setup.py
jdiazdelrio: Changed back to 2.0.1.dev0

requirements.txt
jdiazdelrio: Removed from requirements.txt

jdiazdelrio:
@AndrewAnnex There is an issue with the configuration of conda for Python3.4 in AppVeyor that I cannot figure out.  Could you have a look? Maybe it is something you have seen in the past.  Thanks!

AndrewAnnex:
it looks like the meta.yaml needs to be updated to include the optional dependencies in some way. This is a concern I had with the approach taken here as we are changing the dependencies at runtime which is not easy to handle for package managers. That is why an approach that follows [pep 508 extras](https://www.python.org/dev/peps/pep-0508/#extras) might be better. But idk if conda has an equivalent mechanism.

AndrewAnnex:
although it is unclear why just 3.4 is having the issue right now

Juanlu001:
You can use

```
- dependency  # [py34]
```

for conditional dependencies https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#preprocessing-selectors

AndrewAnnex:
the error appears to be: 

RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://pypi.python.org/simple/certifi/r

jdiazdelrio:
I tried to fix it by adding the certifi dependencies earlier today, but it didn't work.
This will need a bit more investigation.

Jorge

Jorge Diaz del Rio
ODC Space
www.odcspace.com

AndrewAnnex:
@jdiazdelrio I made a branch #224 that I think should have the same effect as this PR but makes the process a bit more explicit. In my PR users with old SSL version will be warned when the installation fails to install spiceypy using an alternative command `pip install spiceypy[security]`. This will install `requests[security]` and follow an alternative path in the download comman.  Requests underneath uses pyopenssl and urllib3 to do what this PR does. I have half tested this using my mac's system python2.7 which shows that it should work

AndrewAnnex:
So I made another pr #225 that I think is a good enough fix for now. The issue causing the 34 python conda build to fail was that certifi was listed as a required dependency (it needed to be moved to the ssl if statement a few lines below), and oddly that version of certifi is not available in the default anaconda channel for 34 (https://anaconda.org/anaconda/certifi/files), So conda could not resolve the dependency. It is available through the conda-forge channel but I don't think any other code changes will be needed other than moving certifi. You can see the diff between your code and mine here: https://github.com/jdiazdelrio/SpiceyPy/compare/master...AndrewAnnex:changes_to_202_jdiazdelrio

jdiazdelrio:
@AndrewAnnex This may work, although if not strictly necessary, I would leave the **certifi** requirement where it was, as it fixed a side issue with the Python3.6 version distributed within www.python.org Mac installer.

In the ReadMe.rtf provided within that installer, the section on "**Certificate verification and OpenSSL**" notes that, as of the 3.6.x installer, the Python supplied no longer links with the deprecated Apple-supplied system OpenSSL libraries but rather supplies a private copy of OpenSSL 1.0.2 which does not automatically access the system default root certificates.  It provides a sample command script included in /Applications/Python 3.6 to install a curated bundle of default root certificates from the third-party certifi package (https://pypi.python.org/pypi/certifi).

AndrewAnnex:
@jdiazdelrio wow this stuff is tricky. I think I need to take a break from SSL issues for a few days. In short this will need to be fixed, probably the way you say in the last paragraph.

AndrewAnnex:
@jdiazdelrio ok so now that I understand it a bit better, I think it could be acceptable to always require certifi. To fix the conda builds we can update the appveyor steps to include using the conda-forge version of certifi (https://github.com/conda-forge/certifi-feedstock) as the july version of certifi is available there.

jdiazdelrio:
@AndrewAnnex The changes have been implemented.  The failures you see on Travis where due an service outage this morning.  If you restart the tests, they will all pass (I have done that on my fork)

AndrewAnnex:
Merged! 🎉

PR #210, opened 2017-10-08
Issue # 209 - dafac (append comments to DAF) not quite there yet

drbitboy:
Hi Andrew,

I did a test by hand, but ran out of time to code a proper test; I should get to that later (probably next week).

Do you have a framework you use run tests that create new kernels?

Brian Carcich

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -1590,7 +1590,7 @@ def dafdc(handle):
     :type handle: int
     """
     handle = ctypes.c_int(handle)
-    libspice.dafcc_c(handle)
+    libspice.dafdc_c(handle)
 
 
 @spiceErrorCheck
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -1515,10 +1515,10 @@ def dafac(handle, n, lenvals, buffer):
     :type buffer: Array of strs
     """
     handle = ctypes.c_int(handle)
-    buffer = stypes.listToCharArrayPtr(buffer)
     n = ctypes.c_int(n)
     lenvals = ctypes.c_int(lenvals)
-    libspice.dafac_c(handle, n, lenvals, ctypes.byref(buffer))
+    buffer = stypes.listToCharArrayPtr(buffer,xLen=lenvals,yLen=n)
+    libspice.dafac_c(handle, n, lenvals, buffer)
 
 
 @spiceErrorCheck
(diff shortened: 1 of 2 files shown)


AndrewAnnex:
take a look at https://github.com/AndrewAnnex/SpiceyPy/blob/master/spiceypy/tests/test_wrapper.py for example tests (this is the framework?). I have a local branch where I attempted to implement a wrapper fro the hidden `dafopn_` function that would be needed to make a proper test but ran into a segfault issue (I did this before for DAS kernels by wrapping `dasonw_`). I can take a look at that branch again and try to get it working for you.

AndrewAnnex:
oops, I think that was for dasonw_ not dafonw_.

AndrewAnnex:
so I looked at tspice (the naif test framework) and noticed that spk kernels are daf kernels, so I think you could reuse some of the code I have for the spkw function tests (like test_spkw08 ) to test this function.

drbitboy:
Yup, I used CK because I was working with CKs earlier today and was familiar with spiceypy.ckw02, but it's the same idea.  Funny that DAFONW (create new DAF) is available to FORTRAN SPICE, but there is not dafonw_c in CSPICE, so there is no spiceypy.dafonw call.

I finished dafac and dafec tests; this pull request supersedes the earlier one:  it includes the same changes, plus the new tests.

drbitboy:
Regarding wrapper for dasonw_ (FORTRAN DAS Open NeW):

Since that is a FORTRAN routine, all visible arguments are pass-by-reference, plus it needs FORTRAN character string lengths passed by value as FORTRAN INTEGERS:

In C:

    /* Subroutine */ int dafonw_(char *fname, char *ftype, integer *nd, integer *
            ni, char *ifname, integer *resv, integer *handle, ftnlen fname_len,
            ftnlen ftype_len, ftnlen ifname_len)

Those last three (type ftnlen) fname_len, ftype_len, and ifname_len, are implicit arguments, as you can see from teh call in FORTRAN:

          ENTRY DAFONW ( FNAME, FTYPE, ND, NI, IFNAME, RESV, HANDLE )
    [...]
    C     CHARACTER*(*)         FNAME
    C     CHARACTER*(*)         FTYPE
    C     INTEGER               ND
    C     INTEGER               NI
    C     CHARACTER*(*)         IFNAME
    C     INTEGER               RESV
    C     INTEGER               HANDLE


The conventions involved (FORTRAN DASONW => dasonw_; how character strings are passed) may be architecture dependent, so getting it working on all platforms could be a headache, although it is F2C so it may be universal.

drbitboy:
+0.07%, woo hoo!

one part in 1400.

AndrewAnnex:
Thanks! Every bit counts! merging.

PR #213, opened 2017-10-09
Misc new and improved tests

drbitboy:
Hi Andrew,

I am confused as to why this cannot be automatically merged; maybe it's the recent additions to tsetyr.

Anyway, here are some new and improved tests.  rdtext is next ...

Brian Carcich

Update:  ah, I failed to check stylistic issues; I'll sort that out and re-issue; this is not your problem!


drbitboy:
instead of
`len(max(buffer,key=len))`
how about
`max(map(len,buffer))`
?

PR #214, opened 2017-10-09
tests for traceback routines, ftncls, tsetyr

drbitboy:
Hi Andrew,

I closed the previous pull request, pulled your latest master, into my branch, and re-committed the changes.

This one should be fine.

Brian Carcich


AndrewAnnex:
Hey Brian,
So this pr has a number of issues besides the failing tests that will need to be resolved before I can merge it. 

Firstly, I didn't test trcoff (aka I don't really support it) because once called it cannot be reversed. So I didn't want it to change how other tests ran unless it can be always be run as the very last test. Reading your pr this now explains the `do_test_trcoff_last` command but I think the same thing could be achieved by placing the test at the end, possibly even after the teardown function or in a separate test file. This would also hopefully get rid of the `setup_trace_set` function, so the global wouldn't be necessary. Alternatively, there is a pytest plugin that could solve this issue cleanly.

It may be more expeditious to break up the pr so that the issues related to testing trcoff are separated from those for ftncls.
 
My other concerns should have probably been expressed in a contributing guide, but I suppose this post could serve as a place to template it out. 

1. SpiceyPy should as much as possible abstract away the underlying ctypes code.
2. Hidden Spice functions (in the c code that end with an `_` should have well made and documented SpiceyPy wrappers if needed for other tests. 

Those are my main concerns with `test_ftncls`. The hidden functions `txtopn_` and `fn2lun_` should have spiceypy wrappers so in the end the test is as pythonic as the rest of the api for consistency. The asserts for spice.failed should also go away as the error checking decorator handles that need, combined with pytests' `with pytests.raises(..)` context manager the bits calling `getmsg` should clean up nicely. In the end the users should not be directly interacting with the ctypes bits like this test currently does.

spiceypy/tests/test_wrapper.py
```diff
@@ -6648,16 +6732,44 @@ def test_trcnam():
...
-    spice.tsetyr(1969)
-    spice.tsetyr(2100)
-    spice.tsetyr(1969)
+    spice.reset()
+    # Expand 2-digit year to full year, typically 4-digit
+    tmp_getyr4 = lambda iy2: int(spice.etcal(spice.tparse('3/3/%02d' % (iy2,),22)[0]).split()[0])
+    # Find current lower bound on the 100 year interval of expansion,
+    # so it can be restored on exit
+    tsetyr_lowerbound = tmp_getyr4(0)
+    for iy2_test in xrange(100):
```
AndrewAnnex: oh and xrange doesn't exist in python 3, so I just use `range` for tests. If you really need the benefits of a generator in python 2.7 there is a python six function that I have used elsewhere in the code.

drbitboy:
Hi Andrew,

Okay, I'll make interfaces to the *_ routines; that will clean things up some as well.

I need spice.failed() to be True once, that is part of the test (that fn2lun_ throws an error is the only way to be sure it FTNCLS does actually close the file).

So I also need to assert (not spice.failed()) to ensure it is False before the step that sets it True.

Other than that I think I can resolve all of the issues.

Brian

drbitboy:
Btw, why remove the n and lenvals arguments from dafac?

I think keeping the interface as identical as possible to the C interface will help users, because the C documentation is where I go to understand things, and having to check a second source for the actual argument list is inconvenient.

drbitboy:
I need spice.failed() to be True once, that is part of the test (that it throws an error is part of the behavior).

So I need to test it and ensure it is False before the step that sets it True.

drbitboy:
Okay, I figured out a way to do it without checking spice.failed()

PR #216, opened 2017-10-10
new tests and (FORTRAN/F2C) support methods

drbitboy:
Once more with feeling ...


drbitboy:
hah!  meant to use vars(spicey.spiceypy.libspice).keys()

PR #217, opened 2017-10-10
Looking for info in Win/MSVC environment

drbitboy:


drbitboy:
Hi Andrew,

The FORTRAN/F2C functions (.cltext_(), .fn2lun_(), .txtopn_(), .writln_()) in the spiceypy.spiceypy.libspice module (cspice.dll) do not seem to be not visible (public?) on Windows/MSVC e.g. see

  https://ci.appveyor.com/project/AndrewAnnex/spiceypy/build/1.1.620/job/ijl2mh69qg3oqwhy

I notice that you have an entry for the FORTRAN/F2C routine .dasonw() in libspicehelper.py:

    libspice.dasonw_.argetypes = [...]

and you make a call to it in spiceypy.dasonw(), although I don't understand why the libspicehelper member is .argetypes and not .argtypes; a google search for "argetypes" and python yields exactly one hit (i.e. SpiceyPy!).

I don't have ready access to a Win/MSVC system.  My working hypothesis is that something mungs (adds or removes?) underscores when making cspice.dll, but I have limited tools to diagnose what is going on.

Have you gone down this road before, and if so could you pass on the magic knowledge, please, Sensei?

drbitboy:
ah, appveyor/cspice.def!

AndrewAnnex:
ah the `argetypes` is just a typo, go ahead and fix that to be `argtypes`, it is part of how ctypes works and is explained in python's documentation.

and you worked out the cspice.def! that is one of the few differences between the mac/linux and windows side of things I worked out a while ago, simply append each new function to that list so that the visual studio stuff will expose those functions.

drbitboy:
woo hoo.  this young Paduan will move on, but I will not forget the Cave.  Yes, the Cave!

drbitboy:
Okay, looks like I got it this time.  Thanks for the help!

drbitboy:
missed a few things, back in a minute or so

AndrewAnnex:
hey, in the future could you reduce the frequency of pushes to this pr? I got a lot of email spam from it and I think one or two people have now unwatched the repo... also closing and reopening the repo does not do much unless it was an accidental click.

I will review this in a few days.

drbitboy:
sure, sorry about that.  got excited about pushing the coverage near 99% and using CI as a metric.

Is there another channel we can use to communicate about SpiceyPy?  dafgsr_c and some of its siblings are next, but some of them are deprecated:  they need the DAF to have the same endian-ess as the architecture of the .dll/.so.  It is unlikely, if not impossible, that dropping those interfaces will break any legacy code.  going into detail here is probably not the best path.

spiceypy/tests/test_wrapper.py
```diff
@@ -6386,6 +6876,12 @@ def test_subpt():
     sep = spice.vsep(point1, point2) * spice.dpr()
     npt.assert_almost_equal(dist, 16.705476097706171)
     npt.assert_almost_equal(sep, 0.15016657506598063)
+    # Iterable ET argument to spice.subpt()
+    point1Alt1Arr = spice.subpt("near point", "earth", [et-20.,et,et+20.], "lt+s", "moon")
+    assert (3,2,) == point1Alt1Arr.shape
+    assert 0. == spice.vnorm(spice.vsub(point1Alt1Arr[1,0],point1))
```
AndrewAnnex: probably worth using pytest.approx(0.0) instead for all floating point comparisons like this.

spiceypy/tests/test_wrapper.py
```diff
@@ -6750,9 +7336,22 @@ def test_unload():
     spice.furnsh(CoreKernels.testMetaKernel)
     # 4 kernels + the meta kernel = 5
     assert spice.ktotal("ALL") == 5
+    # Make list of FURNSHed non-meta-kernels
+    kernelList = [filnam for filnam,filtyp,srcnam,handle in
```
AndrewAnnex: for comprehension is too long, make this a normal nested for loop for readability.

AndrewAnnex:
oops meant to post below first before making above review comments!

The contributor guide will hopefully in the future mitigate the number of changes I am wanting below for future contributions.

First of all there are a lot of small style issues that I want to see fixed but it would be easier to just do that myself. Such issues include most of the new docstrings, naming of the new 'hidden' functions, not using new lines for try catch statements, trailing commas in tuples and missing spaces. I don't think it would be a great use of our time to go through all of them here so I can make those changes if you grant me permission to in the pr, see: https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/ . As a rule of thumb to be put into the contribution guide is that new code should strive to match as closely as possible to the existing prevailing style in the codebase. This is not a hard rule, but I don't see any style differences worth adopting from this pr. 

Second of all, the added bits to skip downloads appear particular to your workflow or only involved for testing the attempt download function and so should be removed. Most of `test_gettestkernels()` test can be kept but it should be within it's own test file, but the top half could be greatly simplified using the six version of reload `from six.moves import reload_module`. 

Third a small thing I noticed were some checks to the spice `failed()` function in some new functions, the spiceErrorCheck decorator takes care of this so they can be removed. Other asserts after the libspice calls are ok...

Fourth the `thing and 1 or 0` ternary statements are completely new to me, I prefer the if else ternary and would like to see that changed. 

This PR is also too large in my opinion so no new changes should be made to it beyond fixing the issues identified above. Since you made the pr from your master branch in your fork it may be worth checking out a new branch with the changes specific to this pr and making a new pr for that branch, then you can continue developing against your master branch independently on other features like the ones we discussed on gitter if you wish (I will add github issues soon to the project to track them outside of this PR). 

I am happy to make the changes I want to see to this pr if that is too much of a time sink, given that I know what everything should look like it is easy for me to do. But I will need the permissions detailed in the first large paragraph above to do that. 

Otherwise I like the additional functionality this pr, the biggest issue are the style by far.

spiceypy/spiceypy.py
```diff
@@ -1742,18 +1774,18 @@ def dafgsr(handle, recno, begin, end):
...
     begin = ctypes.c_int(begin)
     end = ctypes.c_int(end)
-    data = ctypes.c_double()
+    data = stypes.emptyDoubleVector(125)
     found = ctypes.c_bool()
-    libspice.dafgsr_c(handle, recno, begin, end, ctypes.byref(data),
+    libspice.dafgsr_c(handle, recno, begin, end, data,
                       ctypes.byref(found))
-    return data.value, found.value
+    return stypes.cVectorToPython(data)[:end.value+1-begin.value], found.value
```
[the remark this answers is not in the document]
drbitboy: See https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/dafgsr_c.html, specifically https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/dafgsr_c.html#Exceptions

This routine retrieves a vector of double precision values from [beg:end], where beg and end are limited to 1 and 128, respectively.  So if you ask for beg=1 and end=10, or beg=11 and end=20, you only get 10 values, and it puts item beg (one-based) into data[0], and item end (again, one-based) into data[1+end-beg].  So the `data = stypes.emptyDoubleVector(125)` could be `data = stypes.emptyDoubleVector(1+end-beg)`

Now that I look at it again, the data vector should be sized to 128, not to 125; I was assuming the caller would be getting only summary record data, which exclude the first three bytes.  However, there is nothing stopping the caller from retrieving all 128 values.

The dafgsr_c internal C code protects the user from bad beg:end pairs e.g. overflow as in beg=1, end=1280.  Hmm, maybe there should also be some intelligence overflow about in the Python wrapper.

There is also the issue that the beg:end interval is specified in 1-based form; changing that to be begminus1:end in the Python code resulting in begminus1+1:end here would be Pythonic.  However, since NAIF chose not to do that when they wrote the C wrapper, it makes sense to continue that choice for the Python wrapper.  Otherwise, the NAIF CSPICE documentation, which is all the SpiceyPy user has, would become misleading for that SpiceyPy user.

spiceypy/spiceypy.py
```diff
@@ -4275,29 +4314,30 @@ def ekrcec(handle, segno, recno, column, lenout, nelts=3):
...
-    """
+    """ % (_SPICE_EK_EKRCEX_ROOM_DEFAULT,)
     handle = ctypes.c_int(handle)
     segno = ctypes.c_int(segno)
     recno = ctypes.c_int(recno)
     column = stypes.stringToCharP(column)
     lenout = ctypes.c_int(lenout)
     nvals = ctypes.c_int()
-    cvals = stypes.charvector(ndim=nelts, lenvals=lenout)
+    cvals = stypes.charvector(ndim=nelts, lenvals=lenout.value)
```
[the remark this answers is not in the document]
drbitboy: Yes, definitely replace charvector with emptyCharArray.

I prefer leaving the .value in there.  if we pass the lenout variable to stypes.emptyCharArray(), it is changed (back) to an int as xLen internally; that change is not passed back to the lenout variable outsiide the stypes.charvector scope, but I write in so many languages, some of which _would_ change the variable in the outer scop, that I'd rather not have to think about whether such things might happen.  Even Python can change a keyword argument in some cases e.g. lenout.values += 1; okay, that that is not the argument itself, but using .value spares me having to think about that kind of thing, which is the point.

But if you prefer, I can remove the functionally superfluous .values.  Let me know.

spiceypy/tests/test_wrapper.py
drbitboy: Suggest to decline:  if this is not exactly zero, then there is a problem, which is why I explicitly chose [0,0 == ...] (I almost used [... is 0.0;-]).

But still your call; let me know.

spiceypy/spiceypy.py
```diff
@@ -12668,7 +12734,7 @@ def srfxpt(method, target, et, abcorr, obsrvr, dref, dvec):
     if hasattr(et, "__iter__"):
         return numpy.array(
                 [srfxpt(method, target, t, abcorr, obsrvr, dref, dvec) for t in
-                 et])
+                 et]), True  # Need trailing True to spoof @spiceError* above
```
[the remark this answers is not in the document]
drbitboy: (And you left off the "good catch" too ;-)

I spent a lot of time spinning my wheels on this.  As we found out, I didn't know the "decorator" jargon when you mentioned it, but I had already figured out what the @ stuff was doing functionally.

Also, I found this to be an odd result, because I thought the @spiceFoundExceptionThrower decorator checks the type of last item of the return (res[-1]) and should do nothing it that last item is not a [True/False] bool.  So summat odd is happening there - no wait, it _only_ checks type for the found==False case, so this is a bug, and the decorator should use `elif isinstance(found,bool)` instead of just `else`.

Wait a minute:  couldn't we change
`if isinstance(found, bool) and not found:` and `else:`
to simply
`if found is False:` and `elif found is True:`
in the spiceFoundExceptionThrower decorator?

Wouldn't a better (in the sense of more robust) decorator be
```
def spiceFoundExceptionThrower(f)
    @functools.wrapps(f)
    def wrapper(*args,**kwargs):
        res = f(*args,**kwargs):
        if not isinstance(res,tuple):   ### <== the A.R. side of me wants to do this
            return res
        found = res[-1]                 ### <== aack!
        if found is False:
            raise ...
        elif found is True:
            if len(res) == 2:
                return res[0]
            else:
                return res[:-1]
        return res
```
I'm still bothered by the possibility that the highlighted line (cf. [aack!] above) could throw an error (`found = ((None,)+res[-1:])[-1]`) would be more robust), but presumably that is taken care of by the choice of decorated functions.

drbitboy:
Regarding the naming of the new hidden functions:  I think we should leave the trailing underscore in; if we don't, and NAIF publishes C wrappers (e.g. txtopn_c wrapper for txtopn_), then we have already used the name dafonw for the hidden function and can't re-use it without breaking anyone's existing code.

An alternative would be to use _hidden(...) or _f(...); the latter acknowleges that any _-trailing funtion originates from FORTRAN source.

Let me know.

drbitboy:
I am closing this PR and doing `git checkout -b pr217plus/master-with-fixes` as you suggested.

drbitboy:
P.S. I am also going to remove the `Todo: test` comments, since they are obsolete, if that's okay.

drbitboy:
>Second of all, the added bits to skip downloads appear particular to your workflow [...]

All of that was only to get coveralls.io coverage on the last lines in gettestkernels.py.  AFAICT, that `if __name__ ...` clause is not needed by setup.py or anything else (test_wrapper.py calls downloadKernels() on its own), so in fact I think those final lines are actually particular to **_your_** workflow.  A better solution is to remove the last two or three lines of gettestkernels.py, then the whole _raison d'etre_ for the SKIP_DOWLOAD_KERNELS envvar circus goes away; that's what I'll do.

drbitboy:
Umm, does the contributor's guide exist yet?

spiceypy/tests/test_wrapper.py
drbitboy: Switched to this:

    kernelList = []
    for iKernel in range(spice.ktotal("ALL")):
        filnam, filtyp, srcnam, handle = spice.kdata(iKernel, "ALL", 999, 999, 999)
        if filtyp != "META": kernelList.append(filnam)

PR #220, opened 2017-10-22
add support for FreeBSD

murashin:
Added support for FreeBSD on setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,8 @@ __author__ = 'AndrewAnnex'
 
 # Get OS platform
 host_OS = platform.system()
+# Get platform is Unix-like OS or not
+is_unix = host_OS in ("Linux", "Darwin", "FreeBSD")
 # Get current working directory
 root_dir = os.path.dirname(os.path.realpath(__file__))
 # Make the directory path for cspice
@@ -77,7 +79,7 @@ def check_for_spice():
 
 
 def unpack_cspice():
-    if host_OS in ("Linux", "Darwin", "FreeBSD"):
+    if is_unix:
         cspice_lib = os.path.join(lib_dir, ("cspice.lib" if host_OS is "Windows" else "cspice.a"))
         csupport_lib = os.path.join(lib_dir, ("csupport.lib" if host_OS is "Windows" else "csupport.a"))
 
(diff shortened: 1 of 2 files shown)


AndrewAnnex:
hey @ murashin, thanks for the contribution but it looks like a change in this contribution is breaking the rest of the builds, but I haven't really looked into why yet.
I can also see that the check for Mac/Linux/FreeBSD is fairly repetitive, so maybe it would be worth simplifying the code a bit by either making a helper function to do the check ie:
```python
def is_nix():
    return host_OS in ("Linux", "Darwin", "FreeBSD")
```
or you could also just set a variable at the top of the file:
```python
is_nix = host_OS in ("Linux", "Darwin", "FreeBSD")
```

Thanks!

setup.py
```diff
@@ -145,12 +145,13 @@ def build_library():
...
 def move_to_root_directory():
-    sharedLib = 'spice.so' if host_OS == "Linux" or host_OS == "Darwin" else 'cspice.dll'
+    sharedLib = 'spice.so' if host_OS in ("Linux", "Darwin", "FreeBSD") else 'cspice.dll'
     destination = os.path.join(root_dir, 'spiceypy', 'utils', sharedLib)
     if not os.path.isfile(destination):
         target = os.path.join(cspice_dir, 'lib', sharedLib) \
-                if host_OS == "Linux" or host_OS == "Darwin" else \
+                if host_OS in ("Linux", "Darwin", "FreeBSD") else \
                 os.path.join(cspice_dir, 'src', 'cspice', sharedLib)
+        os.makedirs(destination)
```
murashin: This line seems wrong...

murashin:
Thank you for your reply, I think the line 154 caused build failures.
I agree with your idea on creating a helper function or a variable to check Unix like OS or not.
Thanks!

AndrewAnnex:
I think you have this fixed, the red x came from a failed travis macOS build, but they have been having stability issues recently with the macOs builds so it is not a issue with the code. I restarted the build and once it passes I will merge. 

Just as a curiosity which version(s) of FreeBSD are you using? If you can confirm it is working on those versions I would like to add this info to the main readme.

thanks again!

murashin:
I'm using FreeBSD-12 that is CURRENT (development) branch, so not released yet.
SpiceyPy works fine for several months!

Thank you!

AndrewAnnex:
thanks for the contribution!

PR #221, opened 2017-10-22
possible fix to make 3.4 mac builds faster

AndrewAnnex:
it appears they are not using prebuilt wheels


jdiazdelrio:
One issue I found while working on the HTTPS issue is that  the pip from Python3.4 is outdated.  I have forced the update by using pip install -U pip in line 38 and it worked just fine.

AndrewAnnex:
looks like that fixed it, will merge once all checks are there

PR #223, opened 2017-10-23
Reduce test kernel set by over 400MB, down to < 20MB.

drbitboy:
spiceypy/tests/gettestkernels.py

* Replace de421.bsp (16MB) with de405s.bsp (1.4MB; DE-405 short ca. 1997 to 2010).
* DE405S:  Needs to be \<native\>-IEEE for testing, typically LTL-IEEE for Intel.
* DE405S:  Wrote script make_de405s_native.py to convert from BIG-IEEE kernel at JPL/NAIF.
* DE405S:  Integrate make_de405s_native.py script into gettestkernels.py.
* DE405S:  Shift many time-lookup tests into range of de405s.bsp.
* DE405S:  Add step to spiceypy/tests/gettestkernels.py to convert BIG-IEEE de405s.bsp from NAIF to native byteorder.
* Shift Cassini test kernel coverage from 2004 to 2013, to use smaller kernels.
* CASSINI/2013:  Change ABCORR to LT from LT+S, to NONE from S.
* Drop MER and MRO kernels; use Phobos instead of MRO/MER; hardcode MER location.
* Retrieve smaller Phobos DSK (59k) than original NAIF DSK (16MB).
* Retrieve smaller, and static, Earth rotation state binary PCK.
* Do not retrieve latest Earth binary PCK; it changes too often.
* Remove MarsKernels class.
* Remove obsolete cleanup and download (get*Kernels) routines and calls.
* Remove remove redundant kernelFile.close().
* Improve readability of global downloadTestKernels() routine.

spiceypy/tests/make_de405s_native.py

* New script to convert de405s.bsp from NAIF to native byte-order.
* Does functionally the same thing as SPICE utility BINGO.
* Downloads NAIF de405s.bsp as de405s_bigendian.bsp.
* Converts de405s_bigendian.bsp to def05s.bsp.
* Asserts with checksum to ensure success.

spiceypy/tests/test_wrapper.py

* Adjust assertion expected values to match returns from new kernels
* Adjust parameters in test_dskw02_dskrb2_dskmi2()
* Remove obsolete cleanup calls
* Replace MGS/Mars tests with Cassini/Enceladus tests; adjust expected values
* test_ckcov():  simplify assertion, improve test
* Change ['%s%f%d' % tuple(...)] format statements to ['{}{}{}'.format(...)
* Use Mercury transit of Sun instead ofr MRO transit of Mars for test_occult().


spiceypy/tests/test_wrapper.py
```diff
@@ -986,7 +990,8 @@ def test_dafgsr():
     spice.kclear()
     # Open DAF
     # N.B. The SPK used must use the LTL-IEEE double byte-ordering and format
-    # This should be de421.bsp from the test kernel set
+    # This should be de405s.bsp from the test kernel set
+    #handle = spice.dafopr(CoreKernels.spk)
```
AndrewAnnex: think this line can be removed

spiceypy/tests/test_wrapper.py
```diff
@@ -6699,30 +6768,36 @@ def test_srfxpt():
...
+    expected_spoint = [-143.56046006834264971985,  202.9004595420923067195,
+                       -27.99454299292458969717]
+    expected_obspos = [329627.25001832831185311079,  557847.97086489037610590458,
+                       -217744.02422016291529871523]
     npt.assert_array_almost_equal(spoint, expected_spoint)
     npt.assert_array_almost_equal(obspos, expected_obspos)
     # Iterable ET argument:  et-10, et, et+10
-    ets = [et - 10.0, et, et + 10.0]
-    sdtoArr = spice.srfxpt("Ellipsoid", 'Mars', ets, "LT+S", "MGS", frame, bsight)
+    ets = et - 10. + (np.arange(3) * 10.)
```
AndrewAnnex: use `ets = [et - 10.0, et, et + 10.0]` , and the len assert below can be removed

AndrewAnnex:
ok so the BE/ LE thing...  Is there a major BE OS I am forgetting? I guess you could count stuff running on ARM potentially?
 As per my other post about the phobos DSK we could just host the LE kernel on Github instead of converting it. This would be less code to manage.

spiceypy/tests/test_wrapper.py
drbitboy: The assert is just my AR; it can be removed anyway.  But what does it hurt?

spiceypy/tests/gettestkernels.py
```diff
@@ -69,56 +66,16 @@ def cleanup_Cassini_Kernels():
...
-
 class ExtraKernels(object):
     voyagerSclk_url     = "https://naif.jpl.nasa.gov/pub/naif/VOYAGER/kernels/sclk/vg200022.tsc"
     earthTopoTf_url     = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/fk/stations/earth_topo_050714.tf"
     earthStnSpk_url     = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/spk/stations/earthstns_itrf93_050714.bsp"
-    earthGenPck_url     = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_720101_070426.bpc"
-    earthHighPerPck_url = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/earth_latest_high_prec.bpc"
-    phobosDsk_url       = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/dsk/satellites/phobos_3_3.bds"
+    earthHighPerPck_url = "https://naif.jpl.nasa.gov/pub/naif/generic_kernels/pck/a_old_versions/earth_031228_231229_predict.bpc"
+    phobosDsk_url       = "http://www.astro.cornell.edu/~carcich/MISC/phobos_lores_dsk/phobos_lores.bds"
```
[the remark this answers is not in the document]
drbitboy: See PR conversation (soon).

spiceypy/tests/test_wrapper.py
drbitboy: yup.

spiceypy/tests/test_wrapper.py
AndrewAnnex: nothing, it is just not very helpful either

drbitboy: fair enough, it's going away.

drbitboy:
[Combining a few of the comment subjects above into this one comment for brevity (not my strong suit;-)]

Issues:

0.1. Non-https DSK:  meh, as you say; still it would be breaking the CONtRIBUTING guidelines.  I am not sure https does not work it; it might just be a typo.  In any case, does six.urllib2 actually validate the certificates against an actual authority?  Or is it like the commonly-used [--no-check-certificate] option in wget?

0.2. Non-NAIF DSK:  that page is my old Cornell University Astronomy (CUAstro) Dept. Account (carcich@astro.cornell.edu) ~carcich/public_html/, still active five years after I retired.  I agree it is not ideal, but NAIF's DSKs are may be no more stable (see 1.2 below).

0.3. BE/LE thing:  no, there is no BE OS (yes, we got no bananas;-), but CONTRIBUTING.md makes "keeping things platform-neutral" a Responsibility.



Alternatives (two in addition to your three):

1.0. Non-NAIF DSK;  If nothing else, I could ask Tom Shannon, the CUAstro admin, to host it somewhere more permanent in that domain; I am pretty sure he would do that.  Or maybe you have someplace (other than Github; Google Drive?) to host this file; I think that could also use https.  Still, it's ugly.

1.1. Use 10Mb NAIF DSK:  no if at all possible; I really do not want to give back 50% against the fruits of my hard labor.

1.2. Ask NAIF to host it:  been there, done that, was turned down; Boris said "_**the generic_kernels/dsk area is very much in flux**_", so maybe even the current wasteful Phobos test DSK is unstable.  I re-asked about possibility of a_older_versions/ sub-directory; we'll see what he says.

1.3. Use github to host it:  probably the most robust option.

1.3.1 At 20Mb for all of the kernels, why not go all the way and put a .zip and a .tar.gz in a separate SpiceyPy support repo (although compression will probably be minimal on the binary kernels, which make up the bulk of the test kernels).  I think we can use the raw interface to urllib just the file, or use the git Python interface to extract the contents of the support repo (SpiceyPySupport?  SpiceyPyTestKernels?).

1.4. Nat suggested generating kernels on the fly.  We already do that in some cases (test_getfov(); test_gfsntc(); test_srfc2s(); so one could argue that, in effect, the precedent of having a kernel in the SpiceyPy repo is already broken).





2.1. BE/LE:  IIRC, the byte order requirement on the SPICE side comes from deprecated, but backward-compatible, DAF routines that pre-date the introduction of LTL-IEEE/BIG-IEEE in binary kernels, and the introduction of endian-blind SPICE routines.  So a better solution might be to drop the SpiceyPy interfaces to any endian-sensitive deprecated routines, along with their tests, at which point we no longer need to supply at least one native binary test kernel, we can just take whatever byte order NAIF supplies.

AndrewAnnex:
after reading the above:

1. Seems like making another github repo just for the two test kernels is the best option
2. By platform neutrality I meant macOS/Windows/Linux compatibility. Like using having forward slashes vs back slashes for file paths and the like. BE OS support is not important for the project at the moment. 

I can make a public repo for the two kernels soon. An extra advantage is that the repo would be clone-able so we could have backups hosted by gitlab/bitbucket.

drbitboy:
I agree with using Github for those test kernels, but if we do that then we put **_all_** of the kernels into the github repo:  once we have a single test kernel on Github, it makes no sense to bother with another domain (NAIF/JPL) which we cannot control.

drbitboy:
https://github.com/drbitboy/SpiceyPyTestKernels

drbitboy:
Okay, it's done, using my test kernel repo for now.  We should get some new checks soon.

Obviously at some point we will move it to AndrewAnnex@github, and I can also shrink the binary kernels some more, but this works now (at least it passes in my [linux+python2|3] environment.

Also, I think it may even be faster (I don't think NAIF/JPL has bandwidth that can compete with github).

AndrewAnnex:
ok great, I can just fork the repo and then we can just update the links. I think less than 20Mb is a huge win so for now we can just stick with that for now.

AndrewAnnex:
Ok @drbitboy, I made the fork and I have a pr for adding the LE de405s.bsp kernel see https://github.com/AndrewAnnex/SpiceyPyTestKernels/pull/1 . Let me know if that is ok, since that conversion script isn't part of the spiceypy codebase you can choose to license it under your name as well, just let me know and I will update that pr before merging.

drbitboy:
Okay, that is great.

So there are a few more changes to make here:
* Remove [assert len(...) == 3] and re-do ets list
* Remove commented line [#handle = ...]
* Re-home those kernel links to @AndrewAnnex 

Wait, are you are putting the make_de405s_native.py script into the kernels repo?  It belongs in the SpiceyPy repo, under tests so will not be confused with the codebase.

AndrewAnnex:
I added the LE kernel in that PR I made against my fork of the test kernel repository (I didn't delete the BE one). So as far as I can see there is no need for it to be in this PR once I merge in that PR (I was waiting for you to take a look at it). This will also simplify a number of line changes in the current pr once I merge that other PR into the master branch. If you are very concerned about being able to run the tests with the BE version of the kernel then we could have an if check in the gettestkernels code that downloads the correct version if needed.  Otherwise, I don't understand why it is critical here.

drbitboy:
I am confused; not sure which pr and which "it" you are referring to in several places in that last comment.

can we deal with this tomorrow?  I have to take a break.

Thanks,

Brian

AndrewAnnex:
sure thing, feel free to wait until later before you read this. I will try to be more clear here:

1. I forked your repository for the Test Kernels (https://github.com/drbitboy/SpiceyPyTestKernels)
2. I made a PR against my fork of the Test Kernels at (https://github.com/AndrewAnnex/SpiceyPyTestKernels/pull/1) that includes the following:
    * `de405s.bsp`, which is the LE (aka PC) version of the `de405s_bigendian.bsp` kernel
    * `make_de405s_native.py`, that I used to make the new `de405s.bsp` kernel
    * Updates to the README to include the md5 hash for the new `de405s.bsp` kernel

what I am suggesting is: 

1. with your OK I will merge that PR (https://github.com/AndrewAnnex/SpiceyPyTestKernels/pull/1) into my fork of SpiceyPyTestKernels.
2. Then we can remove `make_de405s_native.py` from **this** PR.
3. `make_de405s_native.py` would now be located in (https://github.com/AndrewAnnex/SpiceyPyTestKernels)
4. Then we can simplify the code in **this** PR a bit more as we do not need to convert the `de405s_bigendian.bsp` kernel within the tests here anymore, as we can directly download the new `de405s.bsp` kernel.

drbitboy:
If we put both LE and BE de405s SPKs into the SpiceyPyTestKernels repo, the make_de405s_native.py script is no longer needed.

drbitboy:
Okay, it looks like appveyor and travis are or will be soon showing success, with the usual ignoreable travis osx/py2.7 hiccup.

The script [make_de405s_native.py] is obsolete and gone, and I am pretty sure all review issues have been addressed.

When you re-pull or re-fork the test kernels repo, then you can change all seventeen occurrences of  [drbitboy] to [AndrewAnnex] in spiceypy/tests/gettestkernels.py and do your internal PR, or just push.

We _could_ make the repo owner an envvar that defaults to AndrewAnnex _a la_

    ghOwner = os.environ("TEST_KERNELS_REPO_OWNER","AndrewAnnex')
    ...
    blah_url = "https://raw.githubusercontent.com/{}/SpiceyPyTestKernels/master/...".format(ghOwner)

to make modifying the testing easier in the future, but the tests will eventually stabilize, so the slight awkwardness in updating tests without the envvar is not enough to compensate for the loss in code clarity.

AndrewAnnex:
ok great @drbitboy, this looks good so I will merge soon

drbitboy:
https://ci.appveyor.com/project/drbitboy/spiceypy/build/1.0.68/job/88jvt1uweykp7nub

hmmph.

AndrewAnnex:
@drbitboy interesting... but it didn't fail in my appveyor builds? maybe there is a more memory efficient way to unpack the list... although it should be a numpy array at that point I think?

drbitboy:
Yes, very odd.  Also, it was a memory error but not a segfault, so whatever caused the problem was caught and an exception thrown.  Maybe the VM or the VM host ran out of memory.

AndrewAnnex:
I will merge this once I can get the last travis macos build to pass, so don't worry about this for now.

drbitboy:
yes, that's what I figured.

drbitboy:
woo hoo!

PR #224, opened 2017-10-24
WIP Alternative solution for users with old OpenSSL installs

AndrewAnnex:
* uses info from PR #202 from jdiazdelrio
* users can use pip install spiceypy[security]

### DIFF
--- a/getspice.py
+++ b/getspice.py
@@ -59,6 +59,7 @@ import six.moves.urllib as urllib
 import io
 import zipfile
 import subprocess
+import ssl
 
 __author__ = 'AndrewAnnex'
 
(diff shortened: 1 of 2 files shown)


jdiazdelrio:
@AndrewAnnex 

I have tried your solution on my development platforms, by running "**pip install .**" The first time (Mac, Python2.7.14), i got the message:
   - ImportError: No module named six.moves.urllib
After installing six, I got the error message:
  - ImportError: No module named requests
After installing requests, I got the error message:
  - requests.exceptions.SSLError: HTTPSConnectionPool(host='naif.jpl.nasa.gov', port=443): Max retries exceeded with url: /pub/naif/misc/toolkit_N0066/C/MacIntel_OSX_AppleC_64bit/packages/cspice.tar.Z (Caused by SSLError(SSLError(1, u'[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:661)'),))
which is the same I was getting in September with the current released version of SpiceyPy

AndrewAnnex:
very odd, I am going to try clearing some state maybe that is causing some issues. The import error for requests I fixed in the 2nd commit to this pr so maybe you need to pull those changes.

AndrewAnnex:
hmm something non obvious is happening here where pip appears to be skipping the setup_requires parameter which should ensure six is installed, OR, it never gets the chance to make that check as I have to download cspice before I make the call to setup and that depends on six/optionally requirements.

jdiazdelrio:
Yes, it never gets the chance to make the checks, 
that is why I decided to create a class that encapsulates the setup/build/install
process, so that the setup is first called, and then the class is executed.  I still don't
understand why my PR only fails for the Python3.4 :(

AndrewAnnex:
ah. ok I will take a look at that again. it looks like this also partially because #198 didn't resolve the 'build time' dependency issue. Matplotlib has a fairly complicated work around and this is discussed a bit at https://github.com/scikit-learn/scikit-learn/issues/7867,

jdiazdelrio:
The solution I have proposed on my PR resolves the issue you
are having on your PR.  It actually builds/install in all
platforms and passes the pytests.  It also install from a clean
install on my Mac/Linux development environments.  I think it
would be worth checking why it does not manage to build the conda
package for Python 3.4 as that is the only problem we have for
having a fully functional setup file.  Unfortunately, I will not
have time this week...

AndrewAnnex:
@jdiazdelrio now I think I understand your PR a lot better now. I will look into it

PR #229, opened 2017-10-27
Fix bug in test_wrapper.py calls to spice.spksub

drbitboy:
1. spkuds returns [double,double,int,int] for [first,last,begin,end] cf [here](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/spkuds_c.html).
2. spksub wants to see [double,double] for fourth and fifth arguments cf.[here](https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/spksub_c.html), so ints begin and end are not likely to be the right arguments here.
3. Confusion was likely caused by conflating the meanings and the names of local variables begin and end between the documentation for those two routines.


AndrewAnnex:
great catch. merging!

drbitboy:
not really a "catch" on my part - it bombed when I used the single 0.5Mb SPK - but I'll take the credit ;-)

PR #230, opened 2017-10-28
Add a few function wrappers

AndrewAnnex:


drbitboy:
Use spice.furnsh(SPK18) to load the SPK; spice.kclear() is part of the KEEPER framework and may not know about kernels loaded by the spklef() routine, IIRC.

drbitboy:
or, if you use spklef to load, then use spkuef to unload.

AndrewAnnex:
dropping this pr in favor of splitting off spkw18 and ck05 into later work

PR #231, opened 2017-10-29
small changes, should hopefully have better spk14 test

AndrewAnnex:


drbitboy:
I'm not always right, but it does often work out that way (my wife loves it when I say that;-).

Just kidding; it's really nice to see it come together like this.

;-)

Issue #181, opened 2017-02-02
 python3.4 support


sidewalkjoe:
Trying to pip install spiceypy, but get the following error message:

Command "python setup.py egg_info" failed with error code 1

My Python version is 3.4.4. I was running spiceypy 1.0 before, and everything worked fine. 

Thanks in advance, and please let me know if additional information is required.

AndrewAnnex:
sorry to hear that, I do need more information however like the operating system,  but I also need the full error output to further diagnose. To speculate for a moment, if you are running on windows this error might be caused by the fact that I dropped support for python 3.4, so pip can't use the provided wheels and fails during the building phase. The solution here would be to upgrade to 3.5 or 3.6 which shouldn't be difficult. If you are on linux/macOS there might be something wrong with pip based on google search, but I can think of a few other causes here as well so the full error message would help.

I will be preoccupied for the next few hours, just as a fyi, so if you find a solution on your own in that time please provide some details for others who may run into the same issue.

sidewalkjoe:
Indeed running on Win10. Sorry to hear you've dropped support for Python 3.4. No way to install it for this version at all?

AndrewAnnex:
there are a few options, for one you could install spiceypy 1.0.0 and then manually replace the python files in your site-packages to those of 1.1.0. The main issue here is that the c compiler used to compile the cspice.dll must match that of the python runtime (which differs from 2 to 3). Because of the way appveyor works, each build is run in a series so it isn't feasible to support all of the minor versions. Alternatively I may be able to rename and then upload one of the python 3 wheels as a 3.4 wheel. The last option other than updating to a newer python would be for you to setup visual studio and the related command line tools which I would not recommend.

I will investigate my swap solution soon, but that could take me more time to do than it would take you to upgrade to a newer python.

AndrewAnnex:
I found some time and uploaded prebuilt python 3.4 wheels and conda packages as per #183. 
You should now be able to pip or conda install spiceypy.

AndrewAnnex:
I will close this issue tomorrow unless there is something wrong with the artifacts.

Issue #168, opened 2016-10-25
Installing on anaconda Python 3.4.4


vancesteven:
Here are the install command and output:
conda install -c https://conda.anaconda.org/andrewannex spiceypy
Fetching package metadata .........
Solving package specifications: ....

UnsatisfiableError: The following specifications were found to be in conflict:
- python 3.4*
- spiceypy
  Use "conda info <package>" to see the dependencies for each package.

vancesteven:
pip also fails, but apparently for different reasons:
sudo -H pip install spiceypy
Collecting spiceypy
  Downloading spiceypy-1.1.0.tar.gz (213kB)
    100% |████████████████████████████████| 215kB 2.6MB/s 
    Complete output from command python setup.py egg_info:
    
    You have not agreed to the Xcode license agreements, please run 'xcodebuild -license' (for user-level acceptance) or 'sudo xcodebuild -license' (for system-wide acceptance) from within a Terminal window to review and agree to the Xcode license agreements.
    Error: cspice object file extraction failed with exit status: ('17664',)
    Unable to find CSPICE at /private/tmp/pip-build-z3cdopq1/spiceypy/cspice. Attempting to Download CSPICE For you:
    Gathering information...
    SYSTEM:    Mac
    PROCESSOR: i386
    MACHINE:   64bit
    Best option: MacIntel_OSX_AppleC_64bit
    URL:  https://naif.jpl.nasa.gov/pub/naif/misc/toolkit_N0065/C/MacIntel_OSX_AppleC_64bit/packages/cspice.tar.Z
    
    Downloading...
    Attempting to download spice...
    Unpacking... (this may take some time!)
    Host OS: Darwin
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/tmp/pip-build-z3cdopq1/spiceypy/

vancesteven:
pip also fails, but apparently for different reasons:

sudo -H pip install spiceypy
Collecting spiceypy
  Using cached spiceypy-1.1.0.tar.gz
    Complete output from command python setup.py egg_info:
    gcc: warning: couldn’t understand kern.osversion ‘15.6.0
    ld: library not found for -ldylib1.o
    collect2: error: ld returned 1 exit status
    Error: compilation of shared spice.so build exit status: ('256',)
    Unable to find CSPICE at /private/tmp/pip-build-m8yfyywh/spiceypy/cspice. Attempting to Download CSPICE For you:
[...]

PR #232, opened 2017-10-29
Jdiazdelrio pr202 conda fix

AndrewAnnex:


AndrewAnnex:
This may be a problem: https://github.com/conda-forge/selenium-feedstock/issues/16 ie conda-forge is dropping 3.4 support. So I would have to maintain my own certifi deployment potentially or wait for anaconda to update or tie to a lower certifi version that is available through official channels

AndrewAnnex:
@jdiazdelrio I had to try a lot of things in order to get python 34 to builds to work, it appears that the main issue is that anaconda/conda-forge are starting to drop 34, so I had to go to an older (still 2017) release of certifi. Take a look at the diff and let me know what you think

That all being said, this problem is basically solved if we drop 34 support. I have no particular need for 34 support other than I was asked about continuing to support  3.4 earlier in the year #181 and earlier #168. Also after going through all of the trial and error I am sort of interested in possibly making the urllib3 pathway the default behavior to remove the confusion for conda/pip for when pyopenssl and such is needed. I am sure this contradicts opinions I have said before but the code smell from the conda/appveyor configs is starting to get to me.

jdiazdelrio:
@AndrewAnnex It looks fine. Using an older version of certifi when no newer version is available will do the job. 

Regarding dropping 3.4, I would not do it. conda may drop 3.4, but Python3.4 will still be around for about 18 months (mar-2019 it's the end-of-life, I think). Maybe for the next SPICE toolkit release Python3.4 is no longer around and then it would be justifiable. 

About making urrlib3 the default pathway, I would be in favor, at least for the time being, since it's always better doing the same for all supported versions, but that's a programmers choice :)

AndrewAnnex:
@jdiazdelrio ok sounds good. how about you update your pr to match the needed changes I made in this branch (then I will close #232 ). Then I can merge in your PR.

jdiazdelrio:
@AndrewAnnex Alright. I will do that tomorrow. Besides the changes to the AppVeyor configuration, and the changes in the certifi version, would you like that I implement your suggestion of putting urllib3 as default pathway?

Am I missing anything else?

AndrewAnnex:
I think there a few comments unaddressed in #202 but they are all minor. The urllib3 only idea can wait for a bit or I may make the change later.

setup.py
```diff
@@ -21,10 +21,13 @@
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
+from __future__ import print_function
```
AndrewAnnex: [picky] is this import needed?

PR #234, opened 2017-11-02
added context manager to disable found checks, updated docs a bit

AndrewAnnex:
Adds a context manager to temporarily disable the found flag exception generator for certain situations when that is needed. updated docs a bit to give some details


AndrewAnnex:
I might rename to `suspend_found_check` or `no_found_check`, interested in suggestions

AndrewAnnex:
https://github.com/conda/conda/issues/6251

spiceypy/spiceypy.py
```diff
@@ -85,18 +88,67 @@ def spiceFoundExceptionThrower(f):
...
+        with spice.no_found_check():
+            name, found = spice.bodc2n(-999991) # found is false, no exception raised!
+
+    Within the context any spice functions called that normally check the found
+    flags will pass through the check without raising an exception if they are false.
+
+    """
+    config.catch_false_founds = False
+    yield
+    config.catch_false_founds = True
```
drbitboy: Cool beans!

Consider making this more robust by resetting ```config.catch_false_founds``` to the state it had when no_found_check() was called.  That would protect against code that effectively nests the calls to the context manager.

Say, for example, you write a wrapper for something that uses no_found_check(), but I use it within an app where the first thing I do is call no_found_check() again because I am porting a C program that handles found itself.

These are the changes to make:

in config.py:
```
    config.foundsList = []   # Initialization
```
and then in here:
```
    config.foundsList.append(config.catch_false_founds)
    config.catch_false_founds = False
    yield
    config.catch_false_founds = config.foundsList.pop()
```

Also, the found_check_on/_off() functions below could return the pre-call state, in case someone decided to roll their own context management.

spiceypy/config.py
```diff
@@ -0,0 +1,25 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+"""
+
+catch_false_founds = True
```
drbitboy: Cool beans!

Yeah, picking a good name is not easy here.  Maybe a comment here to 'splain that catch is a verb.

spiceypy/tests/test_spiceerrors.py
```diff
@@ -53,3 +53,19 @@ def test_foundErrorChecker():
     with pytest.raises(spice.stypes.SpiceyError):
         spice.bodc2n(-9991)
     spice.reset()
+
+def test_disable_found_catch():
```
drbitboy: Suggestion:  call a spice.kclear() to ensure no PCK has defined -9991.

spiceypy/tests/test_spiceerrors.py
```diff
@@ -53,3 +53,19 @@ def test_foundErrorChecker():
...
+    with spice.no_found_check():
+        name, found = spice.bodc2n(-9991)
+        assert not found
+    spice.reset()
```
drbitboy: Not sure spice.reset() is needed; I don't think this results in spice.failed() being True.

drbitboy:
Slicker'n snot on a doorknob!

spiceypy/spiceypy.py
AndrewAnnex: good idea, but I think the same effect can be achieved with a local variable, ie:
```
    current_catch_state = config.catch_false_founds
    config.catch_false_founds = False
    yield
    config.catch_false_founds = current_catch_state
```

the on/off calls shouldn't change in the way you suggest as they are not context managers, so there really isn't a before and after for them. It may instead make sense to make another context manager to enable the found catching as a counter point to the one above, and to add a simple method to check to return the current state instead

spiceypy/tests/test_spiceerrors.py
```diff
@@ -53,3 +53,55 @@ def test_foundErrorChecker():
...
+    spice.kclear()
+    spice.found_check_off()
+    _recursive_call(100)
+    spice.kclear()
+    spice.found_check_on()
+    _recursive_call(100)
+    spice.kclear()
+
+
+def test_found_chekc():
```
drbitboy: Did you mean
```
def test_found_check():
```
?

drbitboy:
This is great, thanks for incorporating my ideas.

I thought about the local variable approach, but figured the list would be easier to comprehend, plus I am never trusting of the whole global/local thing.  But it is fine, of course, better even, as nesting context managers effectively makes a list of them in the yield state.

I'm not saying the on/off calls are context managers, just that returning the initial state from the on/off calls could be useful.  but I see you added get_found_catch_state(), so that is just as good, functionally; it just takes an extra step for the user.

Do we need a test of nesting the context managers, to ensure the local variables are indeed local (or do we just trust Python;-)?

AndrewAnnex:
I trust python ;)

AndrewAnnex:
https://github.com/appveyor/ci/issues/1897

drbitboy:
woo hoooooo!

PR #237, opened 2017-11-05
update install docs

### LINKED ISSUE #233: Add explainer to docs about installation issues
although #202 improved installation success with bare python installs, it does not totally solve the runtime dependency issue because setuptools falls back onto easy_install for the setup_requires and install_requires dependencies. This means that installations may still fail as easy_installs are a bit brittle for some transitive dependencies and ultimately inefficient for things like numpy as it does not take advantage of wheels either. Ultimately try to patch all of the contingencies will take too much effort, so it would just be better to tell users to install wheel, numpy, and urllib3[secure] in that order before trying to install on a clean python.  

see:
https://github.com/pypa/setuptools/issues/917
https://pip.readthedocs.io/en/1.4.1/cookbook.html#controlling-setup-requires
https://www.python.org/dev/peps/pep-0518/

AndrewAnnex:
closes #233

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -6,7 +6,13 @@ SpiceyPy is currently supported on Mac, Linux, and Windows systems.
 
 .. _installation:
 
-To install simply run::
+If you are new to python, it is a good idea to read a bit about it first `<docs.python-guide.org>`_.
+For new installations of python, it is encouraged to install and or update: pip, setuptools, wheel, numpy, six, certifi::
+
+    pip install -U pip setuptools wheel
+    pip install -U numpy six certifi
+
+Then to install SpiceyPy, simply run::
 
     pip install spiceypy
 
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -32,8 +38,42 @@ Or you can start a python interpreter and try importing SpiceyPy like so:
...
+issues for users, in particular for macOS users, with OpenSSL versions older
+than 1.0.1g. This is because older versions of OpenSSL still distributed in some
+environments which are incompatible with TLS1.2. As of late 2017 SpiceyPy has been updated with
+a strategy that can mitigate this issue on some systems, but it may not be totally reliable due to known deficiencies in setuptools and pip.
+
+Another solution is to configure a new python installation that is linked against a newer version
+of OpenSSL, the easiest way to do this is to install python using homebrew, once this is done spiceypy
+can be installed to this new installation of python (IMHO this is the best option).
+
+If your python distribution was installed from the packages available at python.org an included command
```
jdiazdelrio: This command is only distributed for Python 3.6 (and higher).  For earlier versions, the only way out is the home-brew installation.

Issue #240, opened 2017-11-10
spiceypy/utils/spice.so image not found when trying to import spiceypy from script


jmend327:
Hello, I receive the following error when trying to `import spiceypy`:

`Traceback (most recent call last):
  File "/Users/Owner/spitest.py", line 1, in <module>
    import spiceypy
  File "/usr/local/lib/python2.7/site-packages/spiceypy/__init__.py", line 25, in <module>
    from .spiceypy import *
  File "/usr/local/lib/python2.7/site-packages/spiceypy/spiceypy.py", line 27, in <module>
    from .utils.libspicehelper import libspice
  File "/usr/local/lib/python2.7/site-packages/spiceypy/utils/libspicehelper.py", line 34, in <module>
    libspice = CDLL(sitePath)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ctypes/__init__.py", line 365, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(/usr/local/lib/python2.7/site-packages/spiceypy/utils/spice.so, 6): image not found
`

I installed through `pip install spiceypy`. If it helps, nothing is returned when I type in `which spiceypy` either. Any help would be appreciated, thanks!

AndrewAnnex:
Ah yes I think I ran across this last night, try uninstalling spiceypy and then installing it using the command “pip install spiceypy —no-cache-dir”
 The “which spiceypy” command shouldn’t do anything as spiceypy is just a Python package and does not put any executables in your PATH variable.

Let me know if that works

-Andrew Annex

AndrewAnnex:
to further explain, there is some odd interaction with pip and setuptools from my experience last night, this issue occurred when pip assumes to use bdist_wheel instead of install so it skips building the spice shared library (spice.so). I think this can be fixed by overriding the bdist_wheel behavior or maybe there is a better way to do this.

jmend327:
Oh I see, thanks. I get this error when trying to run `pip install spiceypy —no-cache-dir`:

Invalid requirement: '—no-cache-dir'
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_install.py", line 82, in __init__
    req = Requirement(req)
  File "/usr/local/lib/python2.7/site-packages/pip/_vendor/packaging/requirements.py", line 96, in __init__
[...]

Issue #245, opened 2018-01-01
Installation Issue


ramsinghdayal:
```
import pip
pip.main(['install', 'spiceypy'])
import spiceypy as spice
```
OSError: dlopen(/Users/ram/anaconda2/lib/python2.7/site-packages/spiceypy/utils/spice.so, 6): image not found

Please, How can it be solve?

AndrewAnnex:
see #240 , ie the fix is to run the following in the command line: `pip install spiceypy --no-cache-dir`, of course after you uninstall.

Also it is a little strange to programmatically install using pip within python like you posted, but it should be possible do update the commands you made to include that option.

AndrewAnnex:
also if you are using anaconda as the prompt suggests you should instead install using conda from either https://anaconda.org/andrewannex/spiceypy or https://anaconda.org/conda-forge/spiceypy

Issue #243, opened 2017-12-06
spice.so image not found


dstansby:
I've just tried to install SpiceyPy using the instructions, but trying to import it errors:

```python
import spiceypy
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dstansby/miniconda3/lib/python3.6/site-packages/spiceypy/__init__.py", line 25, in <module>
    from .spiceypy import *
  File "/Users/dstansby/miniconda3/lib/python3.6/site-packages/spiceypy/spiceypy.py", line 27, in <module>
    from .utils.libspicehelper import libspice
  File "/Users/dstansby/miniconda3/lib/python3.6/site-packages/spiceypy/utils/libspicehelper.py", line 34, in <module>
    libspice = CDLL(sitePath)
  File "/Users/dstansby/miniconda3/lib/python3.6/ctypes/__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: dlopen(/Users/dstansby/miniconda3/lib/python3.6/site-packages/spiceypy/utils/spice.so, 6): image not found
```

dstansby:
Woops, duplicate of #240

AndrewAnnex:
There is currently an issue with setup tools that will cause a command to be skipped during install that I haven't got around to fixing, see #240 #242 for the fix. This assumes you installed using pip, as you are using anaconda if you ran the conda install command I think you would have not seen the issue.

Issue #242, opened 2017-12-04
no cspice.dll after successful pip install


CAReeser:
installed on 64 Windows 10 using pip install spiceypy. It appears all goes well, except there is no cspice.dll anywhere to be found and import into VSC or python fails with module not found (see below).

C:\Python27\Lib\site-packages\spiceypy>pip install spiceypy
Collecting spiceypy
  Downloading spiceypy-2.1.0.tar.gz (113kB)
    100% |################################| 122kB 423kB/s
Requirement already satisfied: numpy>=1.12.0 in c:\python27\lib\site-packages (from spiceypy)
Requirement already satisfied: six>=1.9.0 in c:\python27\lib\site-packages (from spiceypy)
Requirement already satisfied: certifi>=2017.1.23 in c:\python27\lib\site-packages (from spiceypy)
Building wheels for collected packages: spiceypy
  Running setup.py bdist_wheel for spiceypy ... done
  Stored in directory: C:\Users\crees\AppData\Local\pip\Cache\wheels\b6\77\3a\7cf4930d4ee32df5e32a8cd6bd31ecb6b66f40769cf4b895b5
Successfully built spiceypy
Installing collected packages: spiceypy
Successfully installed spiceypy-2.1.0

C:\Python27\Lib\site-packages\spiceypy>cd\
C:\>cd python27

C:\Python27>python
Python 2.7.14 (v2.7.14:84471935ed, Sep 16 2017, 20:19:30) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import spiceypy as spice
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python27\lib\site-packages\spiceypy\__init__.py", line 25, in <module>
    from .spiceypy import *
  File "C:\Python27\lib\site-packages\spiceypy\spiceypy.py", line 27, in <module>
    from .utils.libspicehelper import libspice
  File "C:\Python27\lib\site-packages\spiceypy\utils\libspicehelper.py", line 34, in <module>
    libspice = CDLL(sitePath)
  File "C:\Python27\lib\ctypes\__init__.py", line 366, in __init__
    self._handle = _dlopen(self._name, mode)
WindowsError: [Error 126] The specified module could not be found
>>>

AndrewAnnex:
[...]

PR #246, opened 2018-01-02
Fix wheel builds

AndrewAnnex:
should eventually fix issue where pip install fails to build/include the spice shared library due to issues with setuptools/wheel


AndrewAnnex:
should actually fix #240 #242 #243 #245

PR #251, opened 2018-02-06
Add ProxyManager for CSPICE

seignovert:
First of all, thanks for all your efforts to provide SPICE routines to Python. Great job!

I may have missed it but it looks like the download of CSPICE fails behind proxies with `urllib3`.
It may be not the best way to do it but here is a small fix to it based on the detection of `HTTP[S]_PROXY` environment variable to enable the `ProxyManager` instead of `PoolManager`.

> Based on the docs, for [`urllib.request.urlopen`](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen), if the proxy settings are detected (for example, when `a *_proxy` environment variable like `http_proxy` is set), ProxyHandler is default installed and makes sure the requests are handled through the proxy.

### DIFF
--- a/getspice.py
+++ b/getspice.py
@@ -198,6 +198,14 @@ class GetCSPICE(object):
            HTTPS GET call to the NAIF server to download the required CSPICE
            distribution package.
         """
+        # Search proxy in ENV variables
+        proxy = False
+        try:
+            proxy = os.environ['HTTP_PROXY']
+            proxy = os.environ['HTTPS_PROXY']
+        except KeyError:
+            pass
+
         # Use urllib3 (based on PyOpenSSL).
         if ssl.OPENSSL_VERSION < 'OpenSSL 1.0.1g':
             # Force urllib3 to use pyOpenSSL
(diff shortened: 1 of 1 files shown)


AndrewAnnex:
hey thanks for the contribution, this bit of code was added in 361a8a8866 and admittedly this is not something I considered during the review. I will post some comments now...

getspice.py
```diff
@@ -198,6 +198,14 @@ def _download(self):
            HTTPS GET call to the NAIF server to download the required CSPICE
            distribution package.
         """
+        # Search proxy in ENV variables
+        proxy = False
+        try:
+            proxy = os.environ['HTTP_PROXY']
```
AndrewAnnex: so if a https_proxy is set and not an http one, this would just go the the except block, I think it makes more sense to try https first...

getspice.py
```diff
@@ -208,9 +216,15 @@ def _download(self):
             import urllib3
 
             try:
-                # Create a PoolManager
-                https = urllib3.PoolManager(cert_reqs='CERT_REQUIRED',
-                                            ca_certs=certifi.where())
+                if proxy:
+                    # Create a ProxyManager
+                    https = urllib3.ProxyManager(proxy,
```
AndrewAnnex: I think you could just have proxy be a None or a Dict, and use the proxymanager only instead of having the if statement / the else and such

getspice.py
seignovert: You right. I changed it to store `*_proxy` env variables into `proxies` dict.

getspice.py
seignovert: I agree but it looks like neither `ProxyManager` nor `PoolManager` are able to manage `dict` or `None`. Unfortunately the if/else block seems to be require 😞. I also added `https` vs. `http` checks based on the previous `proxies` parsing from env variables (with `https` first).

PR #252, opened 2018-02-07
Fix termpt docstring

MStefko:
Docstring for `termpt` mistakenly contains description and link to `limbpt`. Since the two functions have very similar input parameters and purpose, this is hard to notice and can cause confusion, especially due to different definition of `rolstep`.


AndrewAnnex:
thanks for the fix!

MStefko:
Happy to help, thanks for the great toolkit :)

PR #280, opened 2019-01-19
Getting return values, default values for spkpos

kd7uiy:
The "cover" is really an input/ output, with no need to specify an input. This change will make this work as expected, and returns an output.

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -11555,7 +11555,7 @@ def spkcls(handle):
 
 
 @spiceErrorCheck
-def spkcov(spk, idcode, cover):
+def spkcov(spk, idcode, cover=None):
     """
     Find the coverage window for a specified ephemeris object in a
     specified SPK file.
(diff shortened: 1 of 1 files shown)


AndrewAnnex:
hey I will review this more tomorrow but the unit tests need to be fixed also. Thanks for the contribution!

AndrewAnnex:
ok I see why the tests fail but the fix is fairly easy, just make sure you update cover and remove the calls declaring cover to begin with. ie `cover =  spice.spkcov(CoreKernels.spk, tempObj)`. Once those lines are updated the tests should pass and I will merge!

kd7uiy:
I've set up two tests, one that keeps the same type of convention that the old one had (Pre-initialized cov), and one that passes in no arguments, with an output test as I am using.

Not all of the tests have completed, but they seem to be passing at this point in time.

AndrewAnnex:
@kd7uiy This is looking much better now! However I wonder if the default size of 2 for the spice cell is appropriate. Looking at the naif documentation depending on the query there can be many intervals reported by spkcov, so it is worth bumping that number to 2000 like used in the example code (this is twice the max number of intervals which is 1000). Great work and thanks for the contribution again!

kd7uiy:
Good suggestion, done.

Thanks for all of the work on this tool, it saved me a bunch of time!

AndrewAnnex:
@kd7uiy Thanks! I always like to hear that SpiceyPy is helping folks out. I am planning to cut a new release in the next few weeks so if there is anything else you want updated I am happy to review PRs!

PR #283, opened 2019-01-23
first pass at making cells easier to create, names will be fluid

### LINKED ISSUE #281: idea: make spice cells easier to access
instead of importing the support_types module, spice cells and such should be accessible by `spiceypy.cell_double` or `spiceypy.cell_d` or something like that

AndrewAnnex:
will fix #281,
I want cells to be easier/ nicer on the eyes to call, plan to create new classes for each cell type and set type to be used to make things easier also, but those functions will remain as well as ones in support types. will need a new doc section to explain

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -192,6 +192,21 @@ def get_found_catch_state():
     """
     return config.catch_false_founds
 
+def cell_double(cell_size):
+    return stypes.SPICEDOUBLE_CELL(cell_size)
+
+def cell_int(cell_size):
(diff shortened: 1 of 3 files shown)


AndrewAnnex:
before I merge I need to write a small doc section on how to declare/work with spice cells. Changing functions to depend on the new classes will be a major change, possibly requiring a major version bump.

AndrewAnnex:
decided to leave the impl of setitem to a later time, there are a number of simplifications to the spice cell class I wish to make and there are some questions about what happens when you delete a element in the middle of a sequence for cells. (do you collapse the data or just init everything to zero?)

PR #284, opened 2019-01-26
Convert standard Python datetime to ET

kd7uiy:
Adding functionality to convert Python's datetime to an et directly, bypassing the need to convert to a string in program should datetimes be utilized.

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -11638,16 +11638,10 @@ def spkcov(spk, idcode, cover=None):
     spk = stypes.stringToCharP(spk)
     idcode = ctypes.c_int(idcode)
     if cover is None:
-<<<<<<< HEAD
-        cover=stypes.SPICEDOUBLE_CELL(2000)
-    assert isinstance(cover, stypes.SpiceCell)
-    assert cover.dtype == 1
-=======
         cover = stypes.SPICEDOUBLE_CELL(2000)
     else:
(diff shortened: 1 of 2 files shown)


kd7uiy:
Hopefully that should do the trick, pending updates. I haven't done a whole lot of this kind of open source work, but I'm learning. Still have a few more ideas that I'll likely submit in the next day or two.

kd7uiy:
Missed that in the comments. This should do the trick.

AndrewAnnex:
Oh so I merged but basically there is an equivalent required reading document in the CSPICE version of the documentation. I do this for a number of other places in the documentation. I can fix this later or it can just be a small PR

-Andrew Annex

kd7uiy:
I link to the point where the J2000.0 epoch is described at https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/FORTRAN/req/time.html#The%20J2000%20Epoch . I would think that would be good enough, but maybe I'm just missing something? 

In any case, thanks! Got at least one more coming, hopefully tonight.

AndrewAnnex:
https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/req/time.html#The%20J2000%20Epoch

Just the CSPICE “flavor” of the same document 

-Andrew Annex

kd7uiy:
Ahhh, got it. Will make the change now, thanks!

PR #285, opened 2019-01-28
Iterate over units

kd7uiy:
Allows to submit a list of items to be iterated over for unit conversion at once. 

In addition, I changed one of the documentation files from a FORTRAN version to C version from a previous patch.


kd7uiy:
Sorry about that, I really need to figure out how to do the tests locally, will certainly work on that. I have actually at least tested the test case this time locally, and it works. I think I've mentioned this is the first Python open source project I've submitted anything to, will work on being less annoying:-) 

Interestingly enough, Wolfram Alpha gave a lower precision value, I had to search pretty hard to get an accurate value for 1 AU.

PR #306, opened 2019-10-16
adding tests to the install 

AndrewAnnex:
so that pytest can take advantage of them more easily in some CI environments

should help fix an issue in pr https://github.com/conda-forge/spiceypy-feedstock/pull/16

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
-from setuptools import setup, Command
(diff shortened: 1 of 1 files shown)


AndrewAnnex:
@jessemapel I had avoided it up until this point and it seems that it is a point of discussion. The conda-forge build process is not very flexible in this regard (https://conda-forge.readthedocs.io/en/latest/testing.html). It looks like many projects do this, but if there is a easy fix for this that can be implemented within the conda-forge framework that will work for me but I haven't found an easy fix for this.

jessemapel:
@AndrewAnnex I poked around a bit and it looks like getting Anaconda working with the Travis setup would take significant work and make things much more complex. The test dependencies are small, so keeping the module working with pip and letting conda-forge handle Anaconda seems like a reasonable solution.

AndrewAnnex:
I'm closing this pr in favor of including the tests into the source distribution via the manifest.in file. This won't require a new release as I should just be able to upload the new tarball to pypi

PR #307, opened 2019-10-18
including tests using manifest so they are distributed via sdist

AndrewAnnex:


AndrewAnnex:
this pr replaces #306, test code after this pr will be distributed via the source tarball. This won't require a version bump

PR #313, opened 2019-11-01
Fix equality check for cells

michaelaye:
If I'm correct this might have been a serious bug.

`is` checks for object identity, not equality. 
And if I understand the code correctly, you need to check the dtype's value to be 0, 1, or 2, correct?

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -1560,11 +1560,11 @@ def copy(cell):
     assert isinstance(cell, stypes.SpiceCell)
     # Next line was redundant with [raise NotImpImplementedError] below
     # assert cell.dtype == 0 or cell.dtype == 1 or cell.dtype == 2
-    if cell.dtype is 0:
+    if cell.dtype == 0:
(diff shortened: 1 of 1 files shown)


michaelaye:
I just checked that it weirdly still works, but it's certainly not what you need it to be in the code. So I guess it was working fine.

AndrewAnnex:
so this is a fun article explaining why this works as is: 
https://kate.io/blog/2017/08/22/weird-python-integers/ 
stackoverflow also on why this works: 
https://stackoverflow.com/questions/306313/is-operator-behaves-unexpectedly-with-integers 

so it doesn't hurt, but is probably better to compare values

michaelaye:
Wow, thanks for that article, fascinating. But it also clearly shows why “is” is indeed the wrong thing in general. Just lucky that these were small integers. I wonder where the limit is, when it fails. I don’t the article said it.

AndrewAnnex:
It’s anything above 256, article describes it

-Andrew Annex

PR #338, opened 2020-01-18
Fixing SyntaxWarning in python 3.8

migueldvb:
python 3.8 issues the SyntaxWarning below when the test_wrapper.py code is being compiled.
```
test_wrapper.py:1341: SyntaxWarning: "is" with a literal. Did you mean "=="?
```
This happens when using `is` identity check with literals (BPO issue: https://bugs.python.org/issue34850).

### DIFF
--- a/spiceypy/__init__.py
+++ b/spiceypy/__init__.py
@@ -1,7 +1,7 @@
 """
 The MIT License (MIT)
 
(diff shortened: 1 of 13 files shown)


AndrewAnnex:
thanks!

PR #340, opened 2020-01-19
Shift URLs in docstrings to https

migueldvb:


AndrewAnnex:
thanks! keep them coming ;)

PR #343, opened 2020-01-29
add et2datetime function

johan12345:
There is already a `datetime2et` function to convert Python datetime objects to seconds past J2000 epoch, which was added in #284. However, the reverse function to convert ET to datetime was still missing.

I have implemented this `et2datetime` function, which first uses `et2utc` and then converts the resulting date strings into `datetime`s.

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -41,7 +41,7 @@ from .utils.callbacks import (
 import functools
 import numpy
 from contextlib import contextmanager
-from datetime import datetime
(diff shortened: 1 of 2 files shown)


AndrewAnnex:
@johan12345 thanks for the contribution, let me know if this contribution or your other PR fixes a blocking issue. If it is I'll find some time in the next week or two to make a new release.

johan12345:
Thank you! It's not too urgent as I have already built workarounds for these two things.

PR #344, opened 2020-02-01
Fix internal reference in docs

migueldvb:
Labels that are not placed before a section title need an explicit
title.
```
SpiceyPy/docs/basics.rst:7: WARNING: undefined label: installation (if the link has no caption the label must precede a section header)
```


AndrewAnnex:
great thanks for the contribution, I'm fixing some other sphinx warnings in another pr that will be posted soon.

PR #365, opened 2020-04-22
Better exception ideas

AndrewAnnex:
progress towards better exception handling in SpiceyPy

This pr creates a new exceptions file with lots of new exceptions defined with the following hierarchy: 

1. SpiceyPyException, which thinly wraps the old SpiceyError exception class to maintain backwards compatibility
2. A family of SpiceyPyExceptions mixed in with builtin exceptions like IOError, ValueError, etc for example SpiceyPyIOError
3. A large suite of SpiceExceptions that are sub classes that inherit from the family of SpiceyPyExceptions, ie SpiceFILEOPENFAILED which is a SpiceyPyIOError

Since all the exceptions are based on the older SpiceyError Exception, existing codebases will be able to maintain the use of SpiceyError catches. With the new hierarchy, exception handling can be more granular down to the specific Spice Exceptions that are defined in SPICE.

Before this pr can be merged, I will likely want to have a new section of documentation to explain the above in a tutorial and to attempt to find any missing Spice Errors defined in the cspice codebase that I have not included so far, and any doc strings for those exceptions I can find.


spiceypy/spiceypy.py
```diff
@@ -77,16 +77,14 @@ def check_for_spice_error(f: Optional[Callable]) -> None:
...
-            "traceback": qcktrc(200),
-        }
-        msg = stypes.errorformat.format(**errorparts)
+        short = getmsg("SHORT", 26)
+        explain = getmsg("EXPLAIN", 100).strip()
+        long = getmsg("LONG", 321).strip()
+        traceback = qcktrc(200)
         reset()
-        raise stypes.SpiceyError(msg)
+        raise stypes.dynamically_instantiate_spiceyerror(
```
autumnsault, line 85: Ah, so this is cool. You've gone from making the error just based on `msg` to also using the other parts of the error metadata, including traceback.

spiceypy/tests/test_spiceerrors.py
```diff
@@ -43,7 +43,25 @@ def test_get_spicey_exception():
...
+def test_getSpiceyPyIOError():
+    with pytest.raises(spice.support_types.SpiceyPyIOError):
+        spice.furnsh(os.path.join(cwd, "_null_kernel.txt"))
+    spice.reset()
+
+
+def test_getIOError():
+    with pytest.raises(IOError):
+        spice.furnsh(os.path.join(cwd, "_null_kernel.txt"))
+    spice.reset()
```
autumnsault: These look like they're testing the exact same thing giving different results. Perhaps a comment to explain the differences?

spiceypy/utils/support_types.py
```diff
@@ -66,122 +66,48 @@
...
+    SpiceyPyValueError,
+    exceptions,
+)
 
-{traceback}
 
-================================================================================\
-"""
+# Collection of supporting functions for wrapper functions
+__author__ = "AndrewAnnex"
```
autumnsault, line 87: I was really curious about the `__author__` metadata. I knew Python had this capability but had never really explored it. There's kind of an interesting debate on Stack Overflow about whether one should use these (other than in `setup.py`):

https://stackoverflow.com/questions/1523427/what-is-the-common-header-format-of-python-files#

spiceypy/utils/support_types.py
```diff
@@ -66,122 +66,48 @@
...
 
 
-class SpiceyError(Exception):
-    """
-    SpiceyError wraps CSPICE errors.
-    :type value: str
+def short_to_spiceypy_exception_class(short: str) -> Type[SpiceyError]:
```
autumnsault, line 90: Ah, cool, another piece of Python notation I'd never seen before. Is the annotation needed to make it work?

spiceypy/utils/support_types.py
AndrewAnnex: yeah I think it was actually auto-populated by pycharm years ago when I started the project and I've never bothered to remove it and kept the convention. I think some of these metadata, in particular the version are worth having.

spiceypy/utils/support_types.py
AndrewAnnex: SpiceyPy as of 3.0.0 has switched to using type hints, this allows sphinx to autogenerate type information for the docs and for pycharm to do it's magic.

autumnsault: Oh, how cool.

PR #367, opened 2020-05-02
[WIP] double checking that this is building 64 bit cspice

AndrewAnnex:


AndrewAnnex:
it looks like it was doing the right thing

Issue #349, opened 2020-02-05
allow spice.so to be found from env var


AndrewAnnex:
currently spiceypy will on pip install, download and build spice.so for users but some users are offline and would like to be able to install from source with a user provided env variable that points to the shared library file, getspice or setup.py should be updated to look for some env flag and to skip the steps for building spice, libspice.py would need to updated to point there optionally

jessemapel:
Thoughts on environment variable name? SPICE_LIB_PATH?

AndrewAnnex:
yeah that would probably be sufficient, I think it is more common to point to the directory and then expect there to be a file within the directory rather than defining the full path to the file. Alternatively this env var could point to a folder of cspice that has already been downloaded and then the get spice building scripts could operate there instead of attempting to download the source code for spice.... maybe this is really two different scenarios

jessemapel:
I think you could reasonably handle both cases. First search the base directory for the binary, if you don't find it then try and search like it's a cspice install.

AndrewAnnex:
note some of these issues are intended for GSoC

medley56:
We are trying to solve this problem for a system that (for an undetermined reason) absolutely refuses to download cspice from NAIF during pip install (it has TLS 1.2). 

Instead of pointing an environment variable, I suggest modifying `get_spice.py` to search in a specific location for the cspice source. If it doesn't find it, then it downloads it. In fact, it looks like this could be added as an `else` in `InstallCSpice.check_for_spice()`.

AndrewAnnex:
@medley56 that was a fallback that I was also considering and could be easily implemented alongside this change. At the moment I am pretty busy so I hope that someone would be willing to implement the functionality as a contribution, as it would all be a fairly easy thing to do using pathlib / os.environ .

medley56:
So I just implemented this in a rudimentary way by changing like 3 lines of code. It's not particularly safe or rubust though. 

```
258     def check_for_spice():
259         print("Checking the path", cspice_dir)
260         if os.path.exists(cspice_dir):
261             print("Found spice at {0}. Skipping download of CSPICE.".format(cspice_dir))
262             return True
263         else:
264             raise NotImplementedError
265             message = "Unable to find CSPICE at {0}. Attempting to Download CSPICE For you:".format(
266                 cspice_dir
267             )
268             print(message)
269             # Download cspice using getspice.py
270             GetCSPICE(version="N0066")
271             if not os.path.exists(cspice_dir):
[...]

PR #371, opened 2020-05-19
Allow installation with existing cspice.tar.Z

medley56:
Add support for installing CSPICE from local directory, specified with environment variable SPICEYPY_CSPICE_PATH.

e.g. SPICEYPY_CSPICE_PATH=~/Downloads/cspice.tar.Z python setup.py install

Adresses  #349

### DIFF
--- a/get_spice.py
+++ b/get_spice.py
@@ -81,14 +81,18 @@ import urllib.request
 import urllib.error
 from zipfile import ZipFile
 
+CSPICE_SRC_DIR = "CSPICE_SRC_DIR"
+CSPICE_SHARED_LIB = "CSPICE_SHARED_LIB"
 
 host_OS = platform.system()
+# Check if platform is supported
+os_supported = host_OS in ("Linux", "Darwin", "FreeBSD", "Windows")
 # Get platform is Unix-like OS or not
 is_unix = host_OS in ("Linux", "Darwin", "FreeBSD")
 # Get current working directory
 root_dir = os.path.dirname(os.path.realpath(__file__))
 # Make the directory path for cspice
-cspice_dir = os.path.join(root_dir, "cspice")
+cspice_dir = os.environ.get(CSPICE_SRC_DIR, os.path.join(root_dir, "cspice"))
 # Make the directory path for cspice/lib
 lib_dir = os.path.join(cspice_dir, "lib")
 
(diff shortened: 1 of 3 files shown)


get_spice.py
```diff
@@ -257,19 +257,35 @@ def get_cspice():
...
+            spiceypy_cspice_path = os.path.expanduser(os.environ['SPICEYPY_CSPICE_PATH'])
+            if not os.path.isabs(spiceypy_cspice_path):
+                raise EnvironmentError("spiceypy_cspice_path must be an absolute path. Got {0}".format(
+                    spiceypy_cspice_path))
+            print("Environment variable spiceypy_cspice_path is set. Installing CSPICE from {0}.".format(
+                spiceypy_cspice_path))
+            print("Unpacking cspice.")
+            untar_cmd = "tar xzf {0} -C {1}".format(spiceypy_cspice_path, root_dir)
+            proc = subprocess.Popen(untar_cmd, shell=True)
+            proc.wait()
```
jessemapel, line 273: Using a built in Python unpacking tool seems like a more robust choice here. Maybe https://docs.python.org/3/library/shutil.html#shutil.unpack_archive

AndrewAnnex: so the existing process in get_spice does use subprocess (specifically "gunzip | tar xC ") I forget the exact reason why as it has now been years but I think there is something non standard about the linux/macos distributions. 

While I thank @medley56 for the contribution, I think there are cleaner ways of implementing this functionality without changing the code so much, for example the behavior could potentially be isolated to the download function entirely with a if statement to see if the environment variable is defined and if so and the file exists pass it on as a io.ByteIO stream.

medley56: @jessemapel I chose to use `subprocess` to match the existing behavior. I agree that this is not ideal. Also, I think there are now cleaner ways to use `subprocess` than `Popen`. I think `run` is the suggested usage now for basic cases.

@AndrewAnnex I started off modifying the `_download` and `_unpack` methods of `GetCSPICE` but decided it was not idiomatic since we are not downloading anything. The whole `GetCSPICE` class seemed very clearly directed toward the acquisition of CSPICE from NAIF and didn't feel like the right place for the new logic without renaming a bunch of stuff.

medley56:
"Allow edit by maintainers" is on. You should be able to continue work on this branch.

get_spice.py
jessemapel: Something like

```
try:
    Subprocess.run(["tar", "xzf", spiceypy_cspice_path, "-C", root_dir], check=True, shell=True, text=True)
except CalledProcessError as e:
    print(f"Failed to unpack CSPICE archive with return code [{e.returncode}], and error [{e.stderr}].")
```

medley56:
I'm glad I could at least get the ball rolling. Even if it's now a different ball.

AndrewAnnex:
in all seriousness, thanks again @medley56! sometimes a little push is all that is needed

PR #376, opened 2020-07-15
Windows Linking Capability

TommasoPino:
Adding the .lib file produced in the building process it is possible to other python modules (like mine) to compile correctly including a spiceypy dependency.

### DIFF
--- a/get_spice.py
+++ b/get_spice.py
@@ -379,6 +379,11 @@ class InstallCSpice(object):
             )
             try:
                 shutil.copyfile(shared_lib_path, destination)
+                if not is_unix:
+                    sharedlib_lib = 'cspice.lib'
+                    destination_lib = os.path.join(root_dir, "spiceypy", "utils",sharedlib_lib)
+                    shared_lib_lib_path = os.path.join(cspice_dir, "src", "cspice", sharedlib_lib)
+                    shutil.copyfile(shared_lib_lib_path, destination_lib)
             except BaseException as e:
                 sys.exit(
                     "{0} file not found, what happend?: {1}".format(shared_lib_path, e)


AndrewAnnex:
hey @TommasoPino, thanks for the contribution but I am a bit hesitant to merge this. The way spiceypy compiles a shared library is already a bit weird, and if your project wants to link against the static library for cspice you are probably better off directly using the .lib file contained in the cspice distribution (see https://naif.jpl.nasa.gov/pub/naif/toolkit//C/PC_Windows_VisualC_64bit/packages/dscriptn.txt). Plus, I don't know of many projects directly referencing binary files from other projects in this way. I would be willing to hear any detailed arguments supporting this change though.

TommasoPino:
﻿Hello Andrew,
Actually the inclusion of this particularly .lib is not for a static library inclusion but to handle the dynamic library. I’m writing a python module that use deeply cspice inside c++ code and either from python side.
I correctly compiled my module using only the .lib produce by the .bat of you. And linked the cspice.dll without any effort because it is already inside your distribution (the only trick is to load your module before mine to ensure the presence inside the python pool of the cspice.dll)

I would like to avoid of compile the entire cspice library only to have this really small piece that your module already produce (but it is just ignored)

The entire process linux side is still perfectly working.

I hope to be clear and obviously I understand if you do not want that. I am already thinking some other workaround.

In conclusion I would like to directly thank you for this very useful module.

Have a nice day,
Tommaso

AndrewAnnex:
hey Tommaso, so I think you are better off compiling/downloading cspice yourself potentially by using the offline installation mode spiceypy has (https://spiceypy.readthedocs.io/en/master/installation.html#offline-installation) which lets you provide a shared library for spiceypy to use at install time. Using that you could replicate the compilation side (the .lib file should be usable from the cspice distribution) but for windows I know it is really annoying to get all the visual studio tools working. That being said if you are already writing c++ code I think you would have everything you need. Alternatively you could also use the github actions/azure functionality to build cspice for you so you don't have to worry about compiling locally.

AndrewAnnex:
@TommasoPino I am planning on closing this pull request soon, I know it is relatively easy to support your use case within spiceypy but I don't think it is a good idea to do so in this way, and that your package should just use the offline installation mode so you can handle the compilation of spice yourself.

PR #384, opened 2020-09-09
use datetime.fromisoformat instead of strptime

johan12345:
to improve performance of `et2datetime`.

The performance of the datetime parsing itself is enhanced by a factor of 100:
```python
timeit.timeit('datetime.datetime.strptime("2020-09-09T15:00:00.000", "%Y-%m-%dT%H:%M:%S.%f").replace(tzinfo=datetime.timezone.utc)', setup='import datetime')
>>> 10.533629100000013
timeit.timeit('datetime.datetime.fromisoformat("2020-09-09T15:00:00.000").replace(tzinfo=datetime.timezone.utc)', setup='import datetime')
>>> 1.2155922999999973
timeit.timeit('datetime.datetime.fromisoformat("2020-09-09T15:00:00.000+00:00")', setup='import datetime')
>>> 0.16864590000000135
```

This leads to a total performance improvement of ~60% for `et2datetime`:
```python
# before
timeit.timeit('spice.et2datetime(200)', setup='import spiceypy as spice; spice.furnsh("latest_leapseconds.tls")')
>>> 27.4600037
# after
timeit.timeit('spice.et2datetime(200)', setup='import spiceypy as spice; spice.furnsh("latest_leapseconds.tls")')
>>> 11.986385900000002
```

`datetime.fromisoformat` is only available since Python 3.7, I added a fallback to `strptime` for Python 3.6.

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -13399,6 +13399,12 @@ def datetime2et(dt: Union[Iterable[datetime], datetime]) -> Union[ndarray, float
         return et.value
 
 
+if hasattr(datetime, 'fromisoformat'):
+    fromisoformat = lambda s: datetime.fromisoformat(s + '+00:00')
+else:
+    fromisoformat = lambda s: datetime.strptime(s, "%Y-%m-%dT%H:%M:%S.%f").replace(tzinfo=timezone.utc)
+
+
 @spice_error_check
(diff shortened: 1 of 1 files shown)


AndrewAnnex:
hey @johan12345, thanks for the contribution! I see that pep8 has a problem with using lambdas, once the pr has been updated to use defs instead it will be good to go.

johan12345:
Sure! I have replaced the lambdas with defs now.

PR #408, opened 2021-03-25
Simplifying functions with ncol/nrow params

AndrewAnnex:
Simplifying functions with ncol/nrow params
    
Some of the early functions I implemented for SpiceyPy were the linear
algebra functions which aren't critical. In those I left some of the
argument row/col dimension arguments which SpiceyPy can figure out for
the user. First pass at fixing some of those and other type annotation
issues I saw.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ spiceypy/utils/__pycache__/
 *.whl
 .pytest_cache/
 *.sqlite3
+/badhashkernel.txt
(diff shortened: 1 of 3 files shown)


AndrewAnnex:
so merging this would cause a breaking api change. Not that this is a bad thing per say but I might take some time before merging this to fix other smaller issues to make things easier for me.

AndrewAnnex:
If I update this pr to make those parameters Options instead, then I can avoid a breaking api change while gaining the functionality and just raise depecrationwarnings for when the user....

AndrewAnnex:
I am going to merge this pr as a lot of the diffs deal with the type annotations less so the actual changes. may need to go back and use deprecated size parameters if provided

PR #410, opened 2021-05-05
definition of locking decorator

TommasoPino:
In order to safely use the spice library in a multithread pool process, it is necessary to lock access to the spice resource.

This is tested with 10000 calls with 16 concurrent threads.

### DIFF
--- a/spiceypy/spiceypy.py
+++ b/spiceypy/spiceypy.py
@@ -74,6 +74,33 @@ _default_len_out = 256
 _SPICE_EK_MAXQSEL = 100  # Twice the 50 in gcc-linux-64
 _SPICE_EK_EKRCEX_ROOM_DEFAULT = 100  # Enough?
 
+import threading
+
+spicelock = threading.Lock()
+
+def spicelock_for_multithread(f):
+    """
+    :return:
+    """
+
+    @functools.wraps(f)
+    def lock(*args, **kwargs):
+
+        if not spicelock.locked():
+            spicelock.acquire()
+            try:
+                # this try handle if some error occurs
+                res = f(*args, **kwargs)
+            finally:
+                if spicelock.locked():
+                    spicelock.release()
+        else:
+            # some library methods calls other library methods, this "else" handle that
+            res = f(*args, **kwargs)
+
+        return res
+
+    return lock
 
 def check_for_spice_error(f: Optional[Callable]) -> None:
     """
@@ -258,6 +285,7 @@ def cell_time(cell_size) -> SpiceCell:
 # A
 
 
+@spicelock_for_multithread
 @spice_error_check
 def appndc(
     item: Union[str, Iterable[str], ndarray, str_],
@@ -280,6 +308,7 @@ def appndc(
         libspice.appndc_c(item, cell)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def appndd(
     item: Union[float, Iterable[float]], cell: Union[SpiceCell, Cell_Double]
@@ -301,6 +330,7 @@ def appndd(
         libspice.appndd_c(item, cell)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def appndi(item: Union[Iterable[int], int], cell: Union[SpiceCell, Cell_Int]) -> None:
     """
@@ -320,6 +350,7 @@ def appndi(item: Union[Iterable[int], int], cell: Union[SpiceCell, Cell_Int]) ->
         libspice.appndi_c(item, cell)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def axisar(axis: Union[ndarray, Iterable[float]], angle: float) -> ndarray:
     """
@@ -343,6 +374,7 @@ def axisar(axis: Union[ndarray, Iterable[float]], angle: float) -> ndarray:
 # B
 
 
+@spicelock_for_multithread
 @spice_error_check
 def b1900() -> float:
     """
@@ -355,6 +387,7 @@ def b1900() -> float:
     return libspice.b1900_c()
 
 
+@spicelock_for_multithread
 @spice_error_check
 def b1950() -> float:
     """
@@ -367,6 +400,7 @@ def b1950() -> float:
     return libspice.b1950_c()
 
 
+@spicelock_for_multithread
 @spice_error_check
 def badkpv(
     caller: str, name: str, comp: str, insize: int, divby: int, intype: str
@@ -394,6 +428,7 @@ def badkpv(
     return bool(libspice.badkpv_c(caller, name, comp, insize, divby, intype))
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bltfrm(frmcls: int, out_cell: Optional[SpiceCell] = None) -> SpiceCell:
     """
@@ -413,6 +448,7 @@ def bltfrm(frmcls: int, out_cell: Optional[SpiceCell] = None) -> SpiceCell:
     return out_cell
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bodeul(body: int, et: float) -> Tuple[float, float, float, float]:
     """
@@ -447,6 +483,7 @@ def bodeul(body: int, et: float) -> Tuple[float, float, float, float]:
     return ra.value, dec.value, w.value, lam.value
 
 
+@spicelock_for_multithread
 @spice_error_check
 @spice_found_exception_thrower
 def bodc2n(code: int, lenout: int = _default_len_out) -> Tuple[str, bool]:
@@ -468,6 +505,7 @@ def bodc2n(code: int, lenout: int = _default_len_out) -> Tuple[str, bool]:
     return stypes.to_python_string(name), bool(found.value)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bodc2s(code: int, lenout: int = _default_len_out) -> str:
     """
@@ -488,6 +526,7 @@ def bodc2s(code: int, lenout: int = _default_len_out) -> str:
     return stypes.to_python_string(name)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def boddef(name: str, code: int) -> None:
     """
@@ -504,6 +543,7 @@ def boddef(name: str, code: int) -> None:
     libspice.boddef_c(name, code)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bodfnd(body: int, item: str) -> bool:
     """
@@ -521,6 +561,7 @@ def bodfnd(body: int, item: str) -> bool:
     return bool(libspice.bodfnd_c(body, item))
 
 
+@spicelock_for_multithread
 @spice_error_check
 @spice_found_exception_thrower
 def bodn2c(name: str) -> Tuple[int, bool]:
@@ -540,6 +581,7 @@ def bodn2c(name: str) -> Tuple[int, bool]:
     return code.value, bool(found.value)
 
 
+@spicelock_for_multithread
 @spice_error_check
 @spice_found_exception_thrower
 def bods2c(name: str) -> Tuple[int, bool]:
@@ -558,6 +600,7 @@ def bods2c(name: str) -> Tuple[int, bool]:
     return code.value, bool(found.value)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bodvar(body: int, item: str, dim: int) -> ndarray:
     """
@@ -584,6 +627,7 @@ def bodvar(body: int, item: str, dim: int) -> ndarray:
     return stypes.c_vector_to_python(values)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bodvcd(bodyid: int, item: str, maxn: int) -> Tuple[int, ndarray]:
     """
@@ -609,6 +653,7 @@ def bodvcd(bodyid: int, item: str, maxn: int) -> Tuple[int, ndarray]:
     return dim.value, stypes.c_vector_to_python(values)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bodvrd(bodynm: str, item: str, maxn: int) -> Tuple[int, ndarray]:
     """
@@ -633,6 +678,7 @@ def bodvrd(bodynm: str, item: str, maxn: int) -> Tuple[int, ndarray]:
     return dim.value, stypes.c_vector_to_python(values)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def brcktd(number: float, end1: float, end2: float) -> float:
     """
@@ -654,6 +700,7 @@ def brcktd(number: float, end1: float, end2: float) -> float:
     return libspice.brcktd_c(number, end1, end2)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def brckti(number: int, end1: int, end2: int) -> int:
     """
@@ -675,6 +722,7 @@ def brckti(number: int, end1: int, end2: int) -> int:
     return libspice.brckti_c(number, end1, end2)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bschoc(
     value: Union[str_, str],
@@ -705,6 +753,7 @@ def bschoc(
     return libspice.bschoc_c(value, ndim, lenvals, array, order)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bschoi(value: int, ndim: int, array: Iterable[int], order: Iterable[int]) -> int:
     """
@@ -727,6 +776,7 @@ def bschoi(value: int, ndim: int, array: Iterable[int], order: Iterable[int]) ->
     return libspice.bschoi_c(value, ndim, array, order)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bsrchc(value: str, ndim: int, lenvals: int, array: Iterable[str]) -> int:
     """
@@ -749,6 +799,7 @@ def bsrchc(value: str, ndim: int, lenvals: int, array: Iterable[str]) -> int:
     return libspice.bsrchc_c(value, ndim, lenvals, array)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bsrchd(value: float, ndim: int, array: ndarray) -> int:
     """
@@ -769,6 +820,7 @@ def bsrchd(value: float, ndim: int, array: ndarray) -> int:
     return libspice.bsrchd_c(value, ndim, array)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def bsrchi(value: int, ndim: int, array: ndarray) -> int:
     """
@@ -793,6 +845,7 @@ def bsrchi(value: int, ndim: int, array: ndarray) -> int:
 # C
 
 
+@spicelock_for_multithread
 @spice_error_check
 def card(cell: SpiceCell) -> int:
     """
@@ -807,6 +860,7 @@ def card(cell: SpiceCell) -> int:
     return libspice.card_c(ctypes.byref(cell))
 
 
+@spicelock_for_multithread
 @spice_error_check
 @spice_found_exception_thrower
 def ccifrm(
@@ -850,6 +904,7 @@ def ccifrm(
     )
 
 
+@spicelock_for_multithread
 @spice_error_check
 def cgv2el(
     center: Iterable[float],
@@ -874,6 +929,7 @@ def cgv2el(
     return ellipse
 
 
+@spicelock_for_multithread
 @spice_error_check
 def chbder(
     cp: Iterable[float], degp: int, x2s: Iterable[float], x: float, nderiv: int
@@ -903,6 +959,7 @@ def chbder(
     return stypes.c_vector_to_python(dpdxs)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def chkin(module: str) -> None:
     """
@@ -916,6 +973,7 @@ def chkin(module: str) -> None:
     libspice.chkin_c(module)
 
 
+@spicelock_for_multithread
 @spice_error_check
 def chkout(module: str) -> None:
     """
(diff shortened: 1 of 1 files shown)


AndrewAnnex (changes requested):
Hey @TommasoPino, first of all thanks for your contribution, as Spice is explicitly single threaded this is not something I've thought about adding myself as of yet but it could make things kinder to users trying to use spiceypy in such a manner without impacting single-thread users.  I posted some general review comments that would firstly need to be answered/addressed but I see a bigger issue in that tests are needed and necessary to demonstrate this working. That is, I would like to see one or two of the tests, like for b1900 and spkezr, duplicated and then modified to make multiple threads to illustrate how this can be used practically. I also think a documentation page/mini tutorial describing this / demoing it would be a great addition although I won't make that a requirement before merging. 

A secondary issue is that I am a little out of practice for using/doing multithreading in Python, so there are some questions I have regarding the particular lock used, and if instead spiceypy should use the synchronised decorator available in the wrapt package (http://wrapt.readthedocs.io/en/latest/examples.html). I also wonder a bit about if external users are implementing their own thread locks and monkey patching the library and how this would effect them, maybe it make sense to optionally disable the decorator in the same manner I provide context managers for the foundflags/error raising decorators.

spiceypy/spiceypy.py
```diff
@@ -14049,7 +14573,8 @@ def tkfram(typid: int) -> Tuple[ndarray, int, bool]:
     return stypes.c_matrix_to_numpy(matrix), next_frame.value, bool(found.value)
 
 
-# @spice_error_check
+# @spicelock_for_multithread
```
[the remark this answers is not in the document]
TommasoPino: Sorry, this was an error during the find/replace process.

spiceypy/spiceypy.py
```diff
@@ -74,6 +74,33 @@
 _SPICE_EK_MAXQSEL = 100  # Twice the 50 in gcc-linux-64
 _SPICE_EK_EKRCEX_ROOM_DEFAULT = 100  # Enough?
 
+import threading
+
+spicelock = threading.Lock()
```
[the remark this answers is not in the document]
TommasoPino: I am new in the multithread field and I used your comment to better understand the documentation. Thanks.
The Rlock is better in this case because it handles recursive calls by the current thread. I will substitute that.

spiceypy/spiceypy.py
```diff
@@ -74,6 +74,33 @@
...
+def spicelock_for_multithread(f):
+    """
+    :return:
+    """
+
+    @functools.wraps(f)
+    def lock(*args, **kwargs):
+
+        if not spicelock.locked():
+            spicelock.acquire()
```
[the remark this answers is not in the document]
TommasoPino: with the usage of the RLock this expression could be deeplòy simplified as you suggested with the "with" statement. Thanks

TommasoPino:
First of all thanks for the fast replay and the suggestions. I will produce an example file to test the current version as you requested.
I prefer the usage of a native package to respect an external one in order to keep the dependency to the minimum.
But, I will study it too

Thanks for the suggestions. I will come back when the modification status will be a more mature state. Thanks for your time

jdiazdelrio:
Hi @TommasoPino and @AndrewAnnex. I'm curious about the outcome of this activity :)

The CSPICE library is not thread-safe in itself. For example, CSPICE would not work on a multi-threaded server where the CSPICE library is dynamically loaded and shared among all threads. Imagine a use case where two users are loading/unloading kernels in different threads started at exactly the same time (T), following this sequence:
- User 1 on thread 1 at T+0s loads a kernel, an LSK;
- User 2 on thread 2 at T+ 5s clears the kernel pool;
- User 1 on thread 1 at T+20s calls `str2et` to convert an UTC string to ET.
If this is implemented using CSPICE library using the SPICE Standard error handling settings, thread 1 will produce an error at T+20s upon calling `str2et` since there will be kernels no loaded at the time user 1 is trying to convert from UTC to ET, and terminate the complete application (both threads and main).

TommasoPino:
Hi @jdiazdelrio, the usage I meant for this modification is not for a server-like environment but for guidances database generation that required for different geometries the invocation of the kernels. The guidances are independent of each other then a concurrent computation could be done. Spawn different processes that load a new instance of the spice library could be a solution but cost a lot in performances and retrieving the separated result is a pain. I found in multithreading a good comprimise.
Obviously, the example you pointed out required a completely different approach.

To answer the @AndrewAnnex's questions I prepared an example file that uses the current 4.0.0 version of and testes the functionalities for multithreading showing the issues in spkezr routine. 

For b1900 the issue is not evident because it reads a variable without modifying it.

I found in this [link ](https://stackoverflow.com/a/28654546) an appropriate way to disable the decorator in order to allow some user that has a personal solution for multithreading calls to disable it in their environments (or enable if we decide to disable it by default).

spiceypy/spiceypy.py
```diff
@@ -74,6 +75,37 @@
 _SPICE_EK_MAXQSEL = 100  # Twice the 50 in gcc-linux-64
 _SPICE_EK_EKRCEX_ROOM_DEFAULT = 100  # Enough?
 
+import threading
+
+_spicelock = threading.RLock()
+
+enablethreadinglock = False
```
AndrewAnnex, line 82: move to https://github.com/AndrewAnnex/SpiceyPy/blob/main/spiceypy/config.py

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+from multiprocessing.dummy import Pool as ThreadPool
+
+
+
+# %%
+b1900 = spice.b1900
+tb1900 = spicelock_for_multithread(b1900)
+
+# %%
+print('Test thread pool without locking for b1900')
```
AndrewAnnex, line 49: move to a new test function within https://github.com/AndrewAnnex/SpiceyPy/blob/main/spiceypy/tests/test_wrapper.py

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+
+for k in defKernels:
+    spice.furnsh(os.path.join(pathKernels,k))
+
+# %%
+spkezr = spice.spkezr
+tspkezr = spicelock_for_multithread(spkezr)
+# %%
+# testing the spicelock_for_multithread decorator
+print('Test thread pool without locking for spkezr')
```
AndrewAnnex, line 81: move to a new test function within https://github.com/AndrewAnnex/SpiceyPy/blob/main/spiceypy/tests/test_wrapper.py

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+    print(e)
+
+
+print('Test thread pool with locking for spkezr')
+spice.enablethreadinglock = True
+with ThreadPool() as pool:
+    test_tspkezr = pool.map(lambda x: tspkezr('EARTH',x,'J2000','NONE','SUN'),test_b1900)
+
+# %%
+# testing the wrapt library
```
AndrewAnnex, line 96: the wrapt library test can be removed

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+
+print('Test thread pool with locking for b1900')
+spice.enablethreadinglock = True
+with ThreadPool() as pool:
+    test_tb1900 = pool.map(lambda x: tb1900(), range(10000))
+
+
+
+# %%
+pathKernels = 'path/to/kernels'
```
AndrewAnnex, line 62: See example test functions in move to a new test function within https://github.com/AndrewAnnex/SpiceyPy/blob/main/spiceypy/tests/test_wrapper.py for this, I have other existing tests for spkezr and my own code for managing loading test kernels for the test suite so some of the code here can be simplified by following those examples as models

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+    'earth_070425_370426_predict.bpc',  # ECEF ephemeris 
+    'pck00010.tpc'                      # Other constants
+]
+
+for k in defKernels:
+    spice.furnsh(os.path.join(pathKernels,k))
+
+# %%
+spkezr = spice.spkezr
+tspkezr = spicelock_for_multithread(spkezr)
```
AndrewAnnex, line 78: for the test, the test should call the function as a user would for the api, so assume that spkezr is already decorated.

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+import functools
+import threading
+
+from spiceypy.spiceypy import spkezr
+
+_spicelock = threading.RLock()
+
+enablethreadinglock = False
+
+def spicelock_for_multithread(f):
```
AndrewAnnex, line 15: Test code should not re-implement the code to be tested.

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+            try:
+                res = f(*args, **kwargs)
+                return res
+            except BaseException:
+                raise
+
+    return lock
+
+# %%
+from multiprocessing.dummy import Pool as ThreadPool
```
AndrewAnnex, line 40: I think the non-dummy threadpool should be used instead. Maybe also the regular pool to see what happens for multiple processes vs threads. In the test suite functions initialize the pools within the test functions and assume 2 threads/processes

spiceConcurrentCall_test.py
```diff
@@ -0,0 +1,109 @@
...
+spice.enablethreadinglock = False
+try:
+    with ThreadPool() as pool:
+        test_spkezr = pool.map(lambda x: tspkezr('EARTH',x,'J2000','NONE','SUN'),test_b1900)
+except BaseException as e:
+    print(e)
+
+
+print('Test thread pool with locking for spkezr')
+spice.enablethreadinglock = True
```
AndrewAnnex, line 91: I was actually envisioning a new set of contextmanagers for the enabling/disabling of the use of your new decorator like I have implemented for the "found" flag checking, see: https://github.com/AndrewAnnex/SpiceyPy/blob/main/spiceypy/spiceypy.py#L150. I also think adding trivial getter/setter/checker functions like I have for the other setting is nicer to use than directly interacting with a parameter like done here.

AndrewAnnex (changes requested):
Hey thanks for continuing to work on this PR. I may not be entirely clear in my contribution documentation, but tests need to be integrated into the existing pytest test infrastructure contained in SpiceyPy (https://github.com/AndrewAnnex/SpiceyPy/blob/main/spiceypy/tests/test_wrapper.py or in a separate file as necessary in that module) rather than the way they are presented here. Sorry if that was not clear. I think that basically the first two (or 4 if you consider the lock/nonlock) tests as written can be added as 2/4 test functions in test_wrapper. You can run the tests yourself by running "pytest" within the top spiceypy directory on your machine, assuming you have installed all of the test-dependencies (pytest and pandas IIRC). As I also handle downloading test kernels for other functions, you should be able to copy/paste what you need from there to follow the existing style established by other tests (like clearing the kernel pool before/after each test). As is these tests are just "dead code" and would not be run by the CI infrastructure. If you are not able to spend more time on this I can probably make the changes needed for you.

AndrewAnnex:
@jdiazdelrio Good points on what SPICE can and can't do. My understanding was that this PR wouldn't actually solve the problem of that exact scenario, but would make it safer to say have multiple threads reading from the kernel pool, same as your example, minus user 2 clearing the kernel pool. For multiple users, multiple processes reading and writing, they would need to maintain separate kernel pools as I understand it. However, maybe there is nothing unsafe with multithreading reads from the kernel pool (ie multiple threads calling spkezr, with a main thread maintaining the kernel pool), in which case that may negate the need of this PR. 

do you have more thoughts on whether this would be useful?

jessemapel:
We've seen SPICE errors from concurrent kernel pool reads using bodvar_c, so I don't think it's safe to just lock kernel loading and unloading.

TommasoPino:
Could you provide a working example to test the case and adjust the PR? Thanks in advance.

Tommaso

jdiazdelrio:
@AndrewAnnex, when having concurrency accessing the kernel pool, it's important to know the state of the kernel pool itself. There are some parameters and features of the kernel pool the will be shared among all threads and that may have an impact on status of the loaded data, the status of the kernel pool itself, and even the performance. As an example using `spkezr`, the SPICE SPK (actually, the underlying DAF subsystem) does buffering in order to improve performance: as records are read from the DAF files, they are saved in an internal buffer maintenance by the DAF subsystem. If any part of that record is needed in order to compute a state, the record is returned without accessing the file. If two concurrent threads "compete" for data in the same file but in different records of that file, the final outcome might be way slower than accessing the data sequentially. Note that due to the nature of the the SPK subsystem, two calls that may look independent, e.g. the state of Phobos w.r.t to Mars and the position of the Earth w.r.t. the Sun, may both require to know the location of the Solar System Barycenter (e.g. if aberration corrections are applied). Similar issues may happen when reading CK, binary PCK or DSKs. Another case is accessing DLA or DAF subsystems directly.

Kernel priority is important as well. Since the kernel pool is shared among all threads, any modification to the kernel pool (adding a new kernel, or manually adding new data using "put-pool" routines `pcpool`, `pipool` or `pdpool`) may lead to unexpected results for those threads that do not expect such modifications. 

The error subsystem is something to look into as well. If I'm not mistaken, SpiceyPy uses the underlying CSPICE error subsystem in "RETURN" mode, which means that until `reset` is called, all non "Error free." calls to CSPICE will return immediately, with undefined results, independently of which thread is calling them.

Any call to `dskstl` will have an impact on all threads, no matter if its locked or not, as it changes the DSK tolerance for all subsequent calls to the DSK subsystem.

AndrewAnnex:
@jdiazdelrio thanks for the details. I think, unless I am mistaken, that this addition would actually help address some, but not all, of the situations you describe, although it does not/cannot address the underlying issues with CSPICE, obviously. As I wrote the following, I think there is a bit of a confusion between multi-threading and multi-processing in the comments above (including mine). In short, this PR would help some situations relating to multi-threading but does not/cannot address the larger issue of side-effect-free/deterministic use of spice with multiple process/threads. 

For an example, for the error subsystem, only a single thread at a time would be allowed to access spice, so if an error occurs in one thread, the call to `reset` would occur and safely return because of the blocks for the other threads.  Same with the example of spkezr, as each thread would be blocked from calling spekzr until the lock is lifted. This should then in effect enforce the serial access aspect, as only one thread at a time would be able to interact with spice. It could be that a lock is not actually appropriate to ensure this, and that instead a Semaphore of 1 should be used instead (equivalent to a mutex iirc), but I will need to think about it more/get some examples coded.

I think the kernel priority/kernel pool modification bit remains unsolved, and can not be addressed by any small code contribution. If I write a variable to the pool from one thread, delete it in another, and expect to read it from a third that wouldn't work even with the locks. Currently however, spiceypy provides no guard rails of any kind, so perhaps this PR should continue forward but with a different "scope" to just address the 1-at-a-time aspect of this question. 

For dskstl, yes I think that and other functions that influence the global state this would be the case, but solving that problem would require using multiple processes in some way to maintain independent spice libraries. That is a bigger/different problem/question than the multithreading question that this PR is trying to help. 

Is that how you understand it?

jdiazdelrio:
@AndrewAnnex, your answer goes in line with my understanding.

But my feeling, in aspects like `spkezr`, is that a user solution based on multiple threads to retrieve state data will be slower than a single-threaded one, because of the likely need of underlying buffering. Note that some high-level SPICE APIs feed from similar data: SPK data is needed for some frame transformations if these need to have aberration corrections; SCLK data is needed for CK access; CK and PCK data might be needed to lookup states on body-fixed or spacecraft-based reference frames... this would mean that this solution, for SPICE intensive applications, might slow down the overall process.

Regarding kernel pool modification, users should think about modification in the sense of reassigning a variable to a new value, or extending/reducing the length of an array within the kernel pool. Therefore, this solution does not help with kernel-pool or CSPICE global state modifications, as you point out.

My guess is that this solution is a good safeguard for non-SPICE-intensive multi-thread applications, where the core of the application can be divided among multiple threads, and each of them can occasionally perform read operations on the SPICE system.

As I said, I'm really curious about the outcome of this activity. If it works it'd be a great contribution to SpiceyPy, even if it's only for some uses (in which case, I'd document very well what can and cannot be done when using SpiceyPy in a multi-threaded environment).

TommasoPino:
> @TommasoPino, is there a way to prohibit certain functions, e.g. `furnsh`, to be called from within a multi-thread environment?

Hi @jdiazdelrio , the only way to prevent accessing `furnsh ` or other function that modifies the kernel pool from child threads is to memorize the name of the father thread and giving only to it access to some specific functions.

spiceypy/spiceypy.py
TommasoPino: moved

spiceConcurrentCall_test.py
TommasoPino: moved in the tests, but I am not sure about it

spiceConcurrentCall_test.py
TommasoPino: moved.

spiceConcurrentCall_test.py
TommasoPino: removed

spiceConcurrentCall_test.py
TommasoPino: As said before, I am not sure about the new test function. Please check it for me :)

spiceConcurrentCall_test.py
TommasoPino: done

spiceConcurrentCall_test.py
TommasoPino: removed

spiceConcurrentCall_test.py
TommasoPino: I used the regular ThreadPool in order to remain in the multithread environment.

spiceConcurrentCall_test.py
TommasoPino: I implemented the entire suite of functions to handle the behavior as "found" with the context manager to avoid direct modification of the parameter

TommasoPino:
Hi @AndrewAnnex, I committed the required modifications for testing and the corrections for CI. Could you please approve the running workflow? Thanks

AndrewAnnex (approved):

AndrewAnnex:
apologies @TommasoPino, I've been preoccupied. This PR has improved greatly since it's initial posting! However, there appears to be some test failures that need to be investigated. I also would like to see short/small tests that demonstrate the usage of the context managers to provide code coverage of those codes (see codecov CI report). See the other context manager tests to see an example of this. 

I also think that this feature needs a dedicated documentation page as @jdiazdelrio suggests to clearly explain situations this can/cannot be used for and demonstrate usage of the context managers.

In any case, I plan to do a release of spiceypy soon to capture small changes since the last release, this PR won't be merged until after that release to keep this change isolated given the breadth of the changes.

spiceypy/spiceypy.py
```diff
@@ -77,6 +78,34 @@
...
+def spicelock_for_multithread(f):
+    """
+    Decorator for spiceypy to avoid concurrent calls to cspice lib.
+
+    :return:
+    """
+    @functools.wraps(f)
+    def lock(*args, **kwargs):
+        if config.enable_threading_lock:
+            with _spicelock:
```
AndrewAnnex, line 94: using [contextlib.suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) here instead could remove the if statement and the duplicated code below in the else clause.
I believe this would work:
```python
ctx = _spicelock if config.enable_threading_lock else contextlib.suppress()
with ctx: 
    try:
    ....
```

AndrewAnnex:
@TommasoPino I think that the build failures were due to changes I made to the test code, and that if we re-run the tests they should all pass, although I can't seem to restart them from within GitHub's ui. If you push any commit (can be empty) to should trigger a rebuild.

spiceypy/spiceypy.py
TommasoPino: Thanks, I am not familiar with context manager. It is a good point to increase the readability. Thanks

TommasoPino:
@AndrewAnnex a new commit has been pushed. Waiting for the approval for the running workflow.
Thanks

spiceypy/spiceypy.py
```diff
@@ -77,6 +78,27 @@
 _SPICE_EK_MAXQSEL = 100  # Twice the 50 in gcc-linux-64
 _SPICE_EK_EKRCEX_ROOM_DEFAULT = 100  # Enough?
 
+import threading
```
AndrewAnnex, line 81: move to top of file with other import statements

spiceypy/spiceypy.py
AndrewAnnex: I forgot to mention in the initial post to add a note reminder that once python 3.6 is deprecated the statement could switch to use 'nullcontext' instead which was introduced in python 3.7.  Maybe this post is a sufficient reminder to myself...

spiceypy/tests/test_wrapper.py
```diff
@@ -47,6 +47,22 @@ def setup_module(module):
     download_kernels()
 
 
+def test_threading_lock():
```
AndrewAnnex, line 50: It would be great to add additional tests for the enable/disable methods and no_threading_lock to improve the codcov report, preferably as additional `def test_...` functions, but they don't need to be to as complicated as this test (you could just use a simple spice function like b1900 to avoid loading kernels etc)

AndrewAnnex:
@TommasoPino yay the tests pass, I left some additional comments that are all pretty simple additions/edits to improve the code coverage. I still think a short documentation section is needed, something like https://github.com/AndrewAnnex/SpiceyPy/blob/main/docs/exceptions.rst#not-found-errors as a new '.rst' file dedicated to this addition. A short explanation derived from the discussions about thread vs process safety and spice would also go a long way. I think you could just contribute a first pass version of this and I could improve it as needed in later commits.

michaelaye:
> As an example using `spkezr`, the SPICE SPK system (actually, the underlying DAF subsystem) does buffering in order to improve performance: as records are read from the DAF files, they are saved in an internal buffer maintained by the DAF subsystem.

@jdiazdelrio This sounds like even using Python's multiprocessing that doesn't use threads, it's still impossible to isolate the kernel pool properly, because it still uses the same installed CSPICE library underneath, correct? 
So, the only way to make this safe is to actually use multiple CSPICE installations, IIUC?

PR #416, opened 2021-07-03
start of adding mypy checks to ci builds

AndrewAnnex:
likely very many type errors


AndrewAnnex:
closing this pr for now, as changes would require major re-write/custom AST parsing to refactor the names of variables in all wrapper functions. Not impossible for sure but not a high priority at the moment either

PR #420, opened 2021-07-28
Updated getfat variables size

marcsit:
I found a bug in the getfat_c wrapper, concerning the default length for the typelen argument; it is set to 4, that leads to a three letter truncation and ‘SCLK’ is truncated to ‘SCL’.

The correction updates the lenght of the arclen and typlen variables to 5 to accomodate a 4 character string.

Tests have been updated accordingly.


AndrewAnnex:
thanks @marcsit! looks good and tests passed so I'll go ahead and merge! :shipit:

PR #426, opened 2021-10-10
Switch to extensions

AndrewAnnex:
WIP to demonstrate using distutils extension to build cspice, works for local tests but seems to build python version specific shared libraries. looses some good things and distributes cspice source in sdist which is undesireable


AndrewAnnex:
closing this for now, really the whole python packaging landscape is in flux at the moment so this is premature

PR #439, opened 2022-02-16
start of N67 support

AndrewAnnex:


AndrewAnnex:
the one failing test is from pointing to an n66 shared library, so it's working! just gotta decide which thing to merge first, possibly this one as I will still need to version bump spiceypy for conda-forge things.

PR #443, opened 2022-03-27
Add new n67 funcs 1

AndrewAnnex:
adds: evsgp4, getfvn, trgsep, twovxf, vprojg

tests failing for twovxf due to a kernel difference I believe. 
No real order to the functions I picked, but next PR will prioritize new functionality over the 49 parity functions


src/spiceypy/spiceypy.py
```diff
@@ -5185,6 +5185,29 @@ def ev2lin(et: float, geophs: Sequence[float], elems: Sequence[float]) -> ndarra
...
+    :param elems: Two-line element data
+    :return: Evaluated state
+    """
+    et = ctypes.c_double(et)
+    assert len(geophs) == 8
+    geophs = stypes.to_double_vector(geophs)
+    assert len(elems) == 10
+    elems = stypes.to_double_vector(elems)
+    state = stypes.empty_double_vector(6)
+    libspice.ev2lin_(ctypes.byref(et), geophs, elems, state)
```
jdiazdelrio, line 5207: shouldn't this call be to `libspice.evsgp4_c` instead?

src/spiceypy/spiceypy.py
```diff
@@ -5609,6 +5632,51 @@ def getfov(
     )
 
 
+@spice_error_check
+def getfvn(
+    inst: str,
+    room: int,
+    shapelen: int = _default_len_out,
```
jdiazdelrio, line 5639: I don't think `shapelen` or `framelen` are needed, assuming that `_default_len_out` is enough to fit any shape name or frame name. By providing these input arguments, you're giving the users the option to set an invalid length and get a truncated output.

src/spiceypy/spiceypy.py
```diff
@@ -15048,6 +15198,27 @@ def vproj(a: ndarray, b: ndarray) -> ndarray:
...
+    Find the projection of one vector onto another vector.
+    All vectors are of arbitrary dimension.
+
+    https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/vprojg_c.html
+
+    :param a: The vector to be projected.
+    :param b: The vector onto which a is to be projected.
+    :return: The projection of a onto b.
+    """
+    ndim = len(a)
```
jdiazdelrio, line 15213: I think you need to check that `ndim = len(a) = len(b)` otherwise you'll have issues with the CSPICE code as it expects that both input arrays, and the output array are of the same length.

src/spiceypy/spiceypy.py
```diff
@@ -14242,6 +14310,57 @@ def trcoff() -> None:
...
+    :param shape2: First target body shape.
+    :param frame2: Reference frame of second target.
+    :param obsrvr: Observing body name.
+    :param abcorr: Aberration corrections flag.
+    :return: angular separation in radians.
+    """
+    _et = ctypes.c_double(et)
+    _targ1 = stypes.string_to_char_p(targ1)
+    _shape1 = stypes.string_to_char_p(shape1)
+    _frame1 = stypes.string_to_char_p(frame1)
```
jdiazdelrio, line 14345: `frame1` (`frame2`) are currently not used: the argument exists for future shape models. Maybe it'd be worth setting a default value, e.g. `'NULL'` as indicated in the CSPICE documentation?

Also, the user may input an empty string, which will cause CSPICE to complain as it checks for non-null strings with at least one character (+ the null terminator). Maybe it is worth checking if that's the case, and passing a single blank character string, i.e. if `frame1 = ''` then `_frame1 = ' '`. Same would apply to `frame2`.

jdiazdelrio (commented):
Hi @AndrewAnnex,

I have had a look at your updates, and I have some comments and suggestions.
Hope this helps.

src/spiceypy/spiceypy.py
AndrewAnnex: good catch

src/spiceypy/spiceypy.py
AndrewAnnex: so far my policy has been, for the most part, to allow overrides in this manner. these are default values so this isn't a significant concern, and GIGO always applies, although it could be a good opportunity to use keyword-only args (pep3102) to provide a bit more a guard without preventing user interaction with these parameters

src/spiceypy/spiceypy.py
AndrewAnnex: stylistically I've stopped doing much LPYL in spiceypy wrappers (although many asserts exists) to lean on the cspice error system more. looking at vprojg, it looks like there are no checks in cspice so an assert can be added

src/spiceypy/spiceypy.py
AndrewAnnex: having two positional arguments with default values is hard to do, I'll leave them and just ignore them to always set `_frame` to `NULL` for now. I assume the empty string comment is just for this function so setting to NULL avoids that. If a user does supply an empty string intentionally or not the CSPICE error would be desireable

Issue #450, opened 2022-06-29
TANGPT N67 API wrapper missing


marcsit:
The TANGPT API wrapper is still missing from the new N67 APIs.

Wrapper and wrapper test are being implemented in the fork: https://github.com/marcsit/SpiceyPy 
 
A pull request will follow after the implementation.

AndrewAnnex:
fixed with #451

PR #451, opened 2022-06-29
Added TANGPT wrapper and test.

marcsit:
As described in https://github.com/AndrewAnnex/SpiceyPy/issues/450

### DIFF
--- a/src/spiceypy/spiceypy.py
+++ b/src/spiceypy/spiceypy.py
@@ -14957,34 +14957,34 @@ def tangpt(
      Point on surface nearest to tangent point, Epoch associated with
      correction locus, Vector from observer to surface point `srfpt'.
     """
-    method = stypes.string_to_char_p(method)
-    target = stypes.string_to_char_p(target)
-    et = ctypes.c_double(et)
-    fixref = stypes.string_to_char_p(fixref)
-    abcorr = stypes.string_to_char_p(abcorr)
-    corloc = stypes.string_to_char_p(corloc)
-    obsrvr = stypes.string_to_char_p(obsrvr)
-    dref = stypes.string_to_char_p(dref)
-    dvec = stypes.to_double_vector(dvec)
+    _method = stypes.string_to_char_p(method)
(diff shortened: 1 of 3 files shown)


src/spiceypy/spiceypy.py
```diff
@@ -14917,6 +14917,87 @@ def szpool(name: str) -> Tuple[int, bool]:
 ################################################################################
 # T
 
+@spice_error_check
+def tangpt(
```
AndrewAnnex, line 14921: please add the argtypes for tangpt to libspicehelper.py https://github.com/AndrewAnnex/SpiceyPy/blob/c82a8720746f4ff80747f19d894071eee4f6607c/src/spiceypy/utils/libspicehelper.py#L2268

src/spiceypy/spiceypy.py
```diff
@@ -14917,6 +14917,87 @@ def szpool(name: str) -> Tuple[int, bool]:
...
+                   "SURFACE POINT".
+    :param obsrvr: Name of observing body.
+    :param dref: Reference frame of ray direction vector.
+    :param dvec: Ray direction vector.
+    :return: "Tangent point": point on ray nearest to surface, Altitude of
+     tangent point above surface, Distance of tangent point from observer,
+     Point on surface nearest to tangent point, Epoch associated with
+     correction locus, Vector from observer to surface point `srfpt'.
+    """
+    method = stypes.string_to_char_p(method)
```
AndrewAnnex, line 14960: okay so I haven't reformatted the rest of this file yet but I've developed a new convention where all parameter type conversion should assign to variables prepended by an underscore ie `_method = stypes.string_to_char_p(method)` to help better support mypy type checking (which wants static typing for everything). Repeat for the rest of the parameters below this line

AndrewAnnex (changes requested):
looks great, just need the argtypes added to libspicehelper.py and the update to the function parameter names noted

AndrewAnnex (commented):
fix names inside tangpt

PR #465, opened 2023-06-27
Add CITATION.cff file

### LINKED ISSUE #460: add a CFF file
would replace the zenodo json file

cgobat:
Resolves #460.

Feel free to make updates as you see fit&mdash;I sourced this info from the existing [codemeta.json](../blob/main/joss/codemeta.json) and the JOSS publication itself.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,6 +1,8 @@
 cff-version: 1.2.0
 title: 'SpiceyPy: a Pythonic Wrapper for the SPICE Toolkit'
-message: If you are publishing work that uses SpiceyPy, please cite SpiceyPy and the SPICE toolkit.
+message: If you are publishing work that uses SpiceyPy, please cite SpiceyPy (using its JOSS DOI)
+         as well as the SPICE toolkit. To cite/refer to a specific version of SpiceyPy, use the
+         associated Zenodo DOI.
 type: software
 authors:
   - given-names: Andrew M.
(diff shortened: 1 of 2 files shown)


CITATION.cff
```diff
@@ -0,0 +1,75 @@
+cff-version: 1.2.0
+title: 'SpiceyPy: a Pythonic Wrapper for the SPICE Toolkit'
+message: If you are publishing work that uses SpiceyPy, please cite SpiceyPy and the SPICE toolkit.
```
AndrewAnnex, line 3: should the message be in quotes? also update to say "cite SpiceyPy using the JOSS doi..." or something to the effect of if citing the use of spiceypy, use the joss doi, but if needing to cite a specific version of spiceypy, cite the specific zenodo doi.

CITATION.cff
```diff
@@ -0,0 +1,75 @@
...
+    orcid: 'https://orcid.org/0000-0002-0455-9384'
+  - given-names: Shankar
+    family-names: Kulumani
+    affiliation: Collins Aerospace
+    orcid: 'https://orcid.org/0000-0002-7822-0471'
+  - given-names: Shin-ya
+    family-names: Murakami
+    affiliation: GFD Dennou Club
+    orcid: 'https://orcid.org/0000-0002-7137-4849'
+identifiers:
```
AndrewAnnex, line 66: I wonder if it would be appropriate to add a references section to cite the NAIF here

AndrewAnnex: and I think maybe preferred-citation should be used with the JOSS doi

AndrewAnnex (commented):
thanks for the contribution @cgobat, just a few small questions and a quick request to add https://github.com/marketplace/actions/cff-validator to the github workflows to validate the file where needed.

CITATION.cff
cgobat: > should the message be in quotes?

No, YAML only needs quotes when the string contains disallowed characters, like `:`. I am actually removing a bunch of unnecessary quotes in the updates I will push shortly.

cgobat: Updated the message. Let me know if it's what you were going for.

CITATION.cff
cgobat: Done & done.

cgobat:
Anything else you wanted to see here, @AndrewAnnex?

PR #482, opened 2024-06-19
updating various ci bits, codecov action may be broken due to change …

AndrewAnnex:
…in tokenless uploads. Source-dist install I think failed due to logic in setup.py that overwrites the commands in wheel is installed


AndrewAnnex:
codecov ci reports are currently broken due to changes in their action, but not a huge concern given the high coverage

PR #501, opened 2025-07-16
Fix ckcov null cell #500

### LINKED ISSUE #500: bool(SpiceCell) and ckcov don't interact well
**Describe the bug**
`ckcov()` doesn't work on long C kernels, reporting the following error:
```
Traceback (most recent call last):
  File ".../src/scanplatform.py", line 453, in <module>
    main()
  File ".../src/scanplatform.py", line 439, in main
    ticks,sclkstrs,ets=get_pointing_instances(ScanPlatformCK,sclkid=encounters[encounter][1])
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../src/scanplatform.py", line 56, in get_pointing_instances
    cover=ckcov(ScanPlatformCK,ckids[0],False,"INTERVAL",0.0,"SCLK",cover)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../venv/lib/python3.11/site-packages/spiceypy/spiceypy.py", line 139, in with_errcheck
    check_for_spice_error(f)
  File ".../venv/lib/python3.11/site-packages/spiceypy/spiceypy.py", line 122, in check_for_spice_error
    raise dynamically_instantiate_spiceyerror(
spiceypy.utils.exceptions.SpiceWINDOWEXCESS: 
================================================================================

Toolkit version: CSPICE_N0067

SPICE(WINDOWEXCESS) --
Cardinality of Window Is Too Small to Contain Result of the Requested Operation
An excess of 2 elements could not be accommodated in the output window.

ckcov_c --> CKCOV --> ZZCKCV01 --> WNINSD

================================================================================
```

The kernel in question has about 17000 discrete pointing instances. I carefully prepared a SpiceCelll with 200,000 elements, expecting it to be plenty large, but no matter how large the pre-initialized cell is, the code still fails:

```
    ...
    ckids=ckobj(ScanPlatformCK)
    print(ckids[0])
    cover = SPICEDOUBLE_CELL(200000)
    cover=ckcov(ScanPlatformCK,ckids[0],False,"INTERVAL",0.0,"SCLK",cover)
    print(wncard(cover))
    ...
```

The problem is in the wrapper around ckcov:

```
@spice_error_check
def ckcov(
    ck: str,
    idcode: int,
    needav: bool,
    level: str,
    tol: float,
    timsys: str,
    cover: Optional[SpiceCell] = None,
) -> SpiceCell:
    """
    Find the coverage window for a specified object in a specified CK file.

    https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/ckcov_c.html

    :param ck: Name of CK file.
    :param idcode: ID code of object.
    :param needav: Flag indicating whether angular velocity is needed.
    :param level: Coverage level: (SEGMENT OR INTERVAL)
    :param tol: Tolerance in ticks.
    :param timsys: Time system used to represent coverage.
    :param cover: Window giving coverage for idcode.
    :return: coverage window for a specified object in a specified CK file
    """
    ck = stypes.string_to_char_p(ck)
    idcode = ctypes.c_int(idcode)
    needav = ctypes.c_int(needav)
    level = stypes.string_to_char_p(level)
    tol = ctypes.c_double(tol)
    timsys = stypes.string_to_char_p(timsys)
    if not cover:
        cover = stypes.SPICEDOUBLE_CELL(20000)
    assert isinstance(cover, stypes.SpiceCell)
    assert cover.dtype == 1
    libspice.ckcov_c(ck, idcode, needav, level, tol, timsys, ctypes.byref(cover))
    return cover
```
The `if not cover:` check is intended to check for when cover is passed in None. `cover` is falsy if it's None, but it's also falsy if it's cardinality is zero. I think what is going on is that the support type SpiceCell has a `__len__()` but no `__bool__()` so Python by default takes `__len__()>0` as `__bool__()`. So when passed an empty SpiceCell, no matter how large it is, the `ckcov()` wrapper reallocates a cover of size 20,000 which isn't enough for my use case, presumably because I need 2 elements for each coverage window.

**Expected behavior**
`ckcov()` uses a pre-initialized empty `SpiceCell` and works without exception on kernels where the SpiceCell is large enough to handle the kernel.

**To Reproduce**
* Furnish a kernel with at least 10,000 coverage windows. This sounds excessive, but there are kernels where there is discrete coverage for every image taken during an encounter with over 10,000 images. One such is `https://pds-rings.seti.org/voyager/ck/vg1_jup_version1_type1_iss_sedr.bc` described at `https://pds-rings.seti.org/voyager/ck/`. 
* Get coverage for the given kernel
```
    ScanPlatformCK="/path/to/vg1_jup_version1_type1_iss_sedr.bc"
    ckids=ckobj(ScanPlatformCK)
    print(ckids[0])
    cover = SPICEDOUBLE_CELL(200000)
    cover=ckcov(ScanPlatformCK,ckids[0],False,"INTERVAL",0.0,"SCLK",cover) # This will raise an exception
    print(wncard(cover))
```

**Desktop (please complete the following information):**
 - SpiceyPy Version 6.0.2
 - OS: Ubuntu Linux 22.04
 - OS Architecture 64bit
 - Python Version 3.11.13
 - Python Architecture 64bit

**Additional context**
I have no idea if there are other places where this matters. I feel like the correct solution (I'll make a pull request) is to change the if to 
```
    if cover is None:
        cover = stypes.SPICEDOUBLE_CELL(20000)
```
but it would be a good idea to check other similar code if any.

kwan3217:
This fixes #500, where ckcov doesn't correctly use a pre-allocated SpiceCell.

### DIFF
--- a/src/spiceypy/tests/gettestkernels.py
+++ b/src/spiceypy/tests/gettestkernels.py
@@ -85,6 +85,8 @@ def cleanup_cassini_kernels() -> None:
 class ExtraKernels(object):
     voyagerSclk_url = "https://raw.githubusercontent.com/AndrewAnnex/SpiceyPyTestKernels/main/vg200022.tsc"
     voyagerSclk_md5 = "4bcaf22788efbd86707c4b3c4d63c0c3"
+    v1jCk_url = "https://pds-rings.seti.org/voyager/ck/vg1_jup_version1_type1_iss_sedr.bc"
+    v1jCk_md5 = "92ac30ccb1c1de6149058aef342da593"
     earthTopoTf_url = "https://raw.githubusercontent.com/AndrewAnnex/SpiceyPyTestKernels/main/earth_topo_050714.tf"
     earthTopoTf_md5 = "fbde06c5abc5da969db984bb4ce5e6e0"
     earthStnSpk_url = "https://raw.githubusercontent.com/AndrewAnnex/SpiceyPyTestKernels/main/earthstns_itrf93_050714.bsp"
@@ -108,6 +110,7 @@ class ExtraKernels(object):
     v02swuck_url = "https://raw.githubusercontent.com/AndrewAnnex/SpiceyPyTestKernels/main/vo2_swu_ck2.bc"
     v02swuck_md5 = "f59ef0556dfc63b55465e152f2d6f5a4"
     voyagerSclk = get_path_from_url(voyagerSclk_url)
+    v1jCk = get_path_from_url(v1jCk_url)
     earthTopoTf = get_path_from_url(earthTopoTf_url)
     earthStnSpk = get_path_from_url(earthStnSpk_url)
     earthHighPerPck = get_path_from_url(earthHighPerPck_url)
@@ -123,6 +126,7 @@ class ExtraKernels(object):
 
 def cleanup_extra_kernels() -> None:
     cleanup_file(ExtraKernels.voyagerSclk)
+    cleanup_file(ExtraKernels.v1jCk)
     cleanup_file(ExtraKernels.earthTopoTf)
     cleanup_file(ExtraKernels.earthStnSpk)
     cleanup_file(ExtraKernels.earthHighPerPck)
@@ -256,6 +260,7 @@ def get_standard_kernels() -> None:
 def get_extra_test_kernels() -> None:
     # these are test kernels not included in the standard meta kernel
     get_kernel(ExtraKernels.voyagerSclk_url, ExtraKernels.voyagerSclk_md5)
+    get_kernel(ExtraKernels.v1jCk_url, ExtraKernels.v1jCk_md5)
     get_kernel(ExtraKernels.earthTopoTf_url, ExtraKernels.earthTopoTf_md5)
     get_kernel(ExtraKernels.earthStnSpk_url, ExtraKernels.earthStnSpk_md5)
     get_kernel(ExtraKernels.earthHighPerPck_url, ExtraKernels.earthHighPerPck_md5)
--- a/src/spiceypy/tests/test_wrapper.py
+++ b/src/spiceypy/tests/test_wrapper.py
@@ -547,6 +547,12 @@ def test_ckcov():
     assert [
         [cover[i * 2], cover[i * 2 + 1]] for i in range(spice.wncard(cover))
     ] == expected_intervals
+    
+    
+def test_ckcov2():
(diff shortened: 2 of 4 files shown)


src/spiceypy/tests/gettestkernels.py
```diff
@@ -85,6 +85,8 @@ def cleanup_cassini_kernels() -> None:
 class ExtraKernels(object):
     voyagerSclk_url = "https://raw.githubusercontent.com/AndrewAnnex/SpiceyPyTestKernels/main/vg200022.tsc"
     voyagerSclk_md5 = "4bcaf22788efbd86707c4b3c4d63c0c3"
+    v1jCk_url = "https://pds-rings.seti.org/voyager/ck/vg1_jup_version1_type1_iss_sedr.bc"
```
AndrewAnnex, line 88: could you open a PR against my `SpiceyPyTestKernels` repo to add the file there, assuming it's only a few kb/mb in size..

src/spiceypy/tests/test_wrapper.py
```diff
@@ -547,6 +548,12 @@ def test_ckcov():
     assert [
         [cover[i * 2], cover[i * 2 + 1]] for i in range(spice.wncard(cover))
     ] == expected_intervals
+    
+    
+def test_ckcov2():
+    ckid = spice.ckobj(ExtraKernels.v1jCk)[0]
+    cover = SPICEDOUBLE_CELL(200000)
```
AndrewAnnex, line 555: additional tests in `test_support_types.py` should be added to ensure test failures if the `__bool__` or `__len__` behavior changes, so that the current behavior (or whatever I work out to be correct later) can have visibility.

AndrewAnnex (changes requested):

src/spiceypy/tests/gettestkernels.py
kwan3217: PR added. I won't update the URL here until that one gets accepted.

AndrewAnnex: @kwan3217 the pr was just merged so this is ready to be updated. thanks!

AndrewAnnex (approved):

PR #503, opened 2025-08-12
Update spiceypy.py

thandal:
Fix documentation (see https://naif.jpl.nasa.gov/pub/naif/misc/toolkit_docs_N0067/C/cspice/spkcpt_c.html)


AndrewAnnex:
thanks for the contribution! merging this in

PR #504, opened 2025-08-20
Returning "cover" from pckcov

bryan-harter:
Howdy, it appears that `pckcov` actually doesn't return anything at the moment, so it just always returns `None`. 

This small change just makes it act more like `ckcov` and `spkcov`.

### DIFF
--- a/src/spiceypy/tests/test_wrapper.py
+++ b/src/spiceypy/tests/test_wrapper.py
@@ -6305,6 +6305,10 @@ def test_pckopn_pckw02_pckcls():
 def test_pckcov():
     ids = spice.cell_int(1000)
 
+    # Call pckfrm to populate ids cell
+    spice.pckfrm(ExtraKernels.earthHighPerPck, ids)
+    assert ids[0] == 3000
+
     # Checks for defaults
     cover = spice.pckcov(ExtraKernels.earthHighPerPck, ids[0])
     result = list(cover)
@@ -6314,7 +6318,6 @@ def test_pckcov():
     # Checks for old way, where if cover is pre-set, it should remain set
     cover = spice.cell_double(2000)
     spice.scard(0, cover)
-    spice.pckfrm(ExtraKernels.earthHighPerPck, ids)
     spice.pckcov(ExtraKernels.earthHighPerPck, ids[0], cover)
     result = list(cover)
     expected = [94305664.18380372, 757080064.1838132]


AndrewAnnex:
@bryan-harter it currently modifies the cover spicecell so it does `work` and isn't actually currently supposed to return anything. But it could be considered an oversight in ensuring a pythonic api as it's just modifying the cell in place:

https://github.com/AndrewAnnex/SpiceyPy/blob/7313ec097ed10cc8a450d39daf04c02ebbd3ce2c/src/spiceypy/spiceypy.py#L10320-L10335

see the test that is still working

https://github.com/AndrewAnnex/SpiceyPy/blob/7313ec097ed10cc8a450d39daf04c02ebbd3ce2c/src/spiceypy/tests/test_wrapper.py#L6305-L6313

I will need to think about that and see what other spice cell functions I've implemented do

bryan-harter:
Ahhhh that makes total sense. Yes, that works when we do that. Thanks very much for such a quick reply! 

Whenever I've used spkcov and ckcov from spiceypy before, I've used the object that is returned, rather than retrieving the same cover object that got modified. We were just running into this issue on IMAP, since we were basically doing something like this in our SPICE handling code:

```
if file_extension = 'bc':
    function_to_call = spiceypy.ckcov
elif file_extension = 'bsp':
    function_to_call = spiceypy.spkcov
cover = function_to_call(...)
# Other things with "cover" here
```

But the other day we got some bpc files now for more accurate info of Earth's barycenter, so we went to add:

```
elif file_extension == 'bpc':
    function_to_call = spiceypy.pckcov
cover = function_to_call(...)
# Other things with "cover" here
```

and were surprised when "cover" was always None. But we could easily change our code around to instead get "cover" again from the same object that we put into the function. I didn't realize that it was a mutable object.

AndrewAnnex:
yeah so a lot of cell functions like spkcov were updated at some point to accept the cover as a optional argument, if you don't provide it a spice cell, it allocates a spicecell for you and so must return it. 

I think it could make sense to return cover here though, I am not sure if it makes sense to adapt the behavior above for it however. Could you update the pr to update the return type annotation? 

In any case, this is a minor issue so don't think it's a blocker for your work...

src/spiceypy/spiceypy.py
```diff
@@ -10326,13 +10326,18 @@ def pckcov(pck: str, idcode: int, cover: SpiceCell) -> None:
...
-    :param cover: Window giving coverage in pck for idcode.
+    :param cover: Optional SPICE Window giving coverage in "pck" for "idcode".
+    :return: coverage window for a specified "idcode" in specified "pck" file
     """
     pck = stypes.string_to_char_p(pck)
     idcode = ctypes.c_int(idcode)
-    assert isinstance(cover, stypes.SpiceCell)
-    assert cover.dtype == 1
+    if cover is None:
+        cover = stypes.SPICEDOUBLE_CELL(20000)
```
AndrewAnnex, line 10335: I believe scard needs to be called on the cell, see https://naif.jpl.nasa.gov/pub/naif/misc/toolkit_docs_N0067/C/cspice/pckcov_c.html

src/spiceypy/tests/test_wrapper.py
```diff
@@ -6304,10 +6304,19 @@ def test_pckopn_pckw02_pckcls():
 
 def test_pckcov():
     ids = spice.cell_int(1000)
+    temp_obj = ids[0]
+
+    # Checks for defaults
+    cover = spice.pckcov(ExtraKernels.earthHighPerPck, temp_obj)
+    result = [x for x in cover]
```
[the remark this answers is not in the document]
bryan-harter: Yeah list(cover) works just fine it appears

src/spiceypy/spiceypy.py
AndrewAnnex: interesting... anyways thanks for adding that I feel better about it now heh

src/spiceypy/tests/test_wrapper.py
```diff
@@ -6304,11 +6304,20 @@ def test_pckopn_pckw02_pckcls():
 
 def test_pckcov():
     ids = spice.cell_int(1000)
+    temp_obj = ids[0]
```
AndrewAnnex, line 6307: this line cause the test to fail, I don't think it's necessary so removing it should fix things

PR #554, opened 2026-07-13
Fix a few quirks of test temporary directory management

markmentovai:
- Don’t special-case macOS’s temporary directory in tests
 - Remove exampleKernels.txt during test clean-up
 - Skip redundant filesystem checks during test clean-up


AndrewAnnex:
@markmentovai thanks for the contribution, looks good to me, will merge once ci finishes.

